PR #136, opened 2018-04-18
Conditional dependency btw libavcodec-ffmpeg56 and libavcodec-ffmpeg-extra56

G3rb:
On several Linux version (Mint 18 for example)  the package libavcodec-ffmpeg56 is replaced by libavcodec-ffmpeg-extra56.
To increase the compatibility I added an exception which defines a conditional dependency between these two packages.

### DIFF
--- a/release_scripts/linux/package.sh
+++ b/release_scripts/linux/package.sh
@@ -345,10 +345,15 @@ fi
 
 for ((i=0;i<${#tab[@]};i++)); do
     if [ $i -ne "0" ]; then
-        echo -n ",    " >> "$CONTROL_FILE"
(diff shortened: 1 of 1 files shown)


AgentScarn:
How do you implement this fix from the .deb? Mint 19 won't install due to the missing libavcodec-ffmpeg56 dependency.

G3rb:
This package will not work on Linux mint 19. It's builded for an Ubuntu 1604 base. So you have to re-compile it on a mint 19. I didn't try it for now in Mint 19 but I think you will have the same issue.
If you want you can change the control file to match the new codecs packages but I think it will be a bit tricky and not sure if it works.

PR #140, opened 2018-05-10
Dark Backgroundcolor could be configurable

jenskuespert:
Hi,

I recently began using OpenBoard and so far everything looks good. What I missed was the ability to select another dark background color...

I hope, I got this right and would be glad if you saw this addition fit to integrate into the main-line.

Regards,

-- 
    - Jens -


UtkarshVerma:
I'd also like to see this feature get implemented. Alongwith that, we could also have a way to choose five colors for the pen to suit the specific background.

rheibel:
Could I also ask for this feature. We have a number of teachers who prefer the pastel dark green chalkboard effect offered on some new android interactive TV's. Light Dark and A custom Background colour for indevidual users would be ideal but even just a config file option to change the dark background would be most welcome.

Thanks to @jenskuespert who looks to have done some work on this but I lack the skills and time to build from source on win 10 64bit, so am looking forward to a official release.

jenskuespert:
I worked on this feature a littel bit more. Now the default size of the grid is stored in the preferences, I also added a display to show the current numerical value while dragging the slider.

If there is any interest I'll create a new pull-request based on the current version.

PR #187, opened 2019-01-09
Always correctly show the stylus color.

### LINKED ISSUE #133: colour choice widget initialization
Hi all and thank you for you precious work.
I had to compile OB by myself (cloned current master git) as I manage a Debian sid based 32 bit system for Italian schools.
Everything works very well, but I feel there could be a little problem with initialization of the colour choice widget.
The icons of its toolbar are not set on start (only 5 void little squares are shown). If I change the tool to anything (e.g. pointer) and then back to Pen, icons appear.
I noticed that in UBApplication::exec, setStylusTool is called to force an initial value of UBStylusTool::Pen 
UBDrawingController::drawingController()->setStylusTool((int)UBStylusTool::Pen)
This is done just before UBApplicationController allocation.
Don't know if this is correct (I haven't studied code in deep), but moving the setStylusTool invocation after UBApplicationController allocation, seems to solve the issue.
I would appreciate a comment by you developers.
Regards
Claudio

sunweaver:
Fixes #133.

### DIFF
--- a/src/core/UBApplication.cpp
+++ b/src/core/UBApplication.cpp
@@ -301,8 +301,6 @@ int UBApplication::exec(const QString& pFileToImport)
     webController = new UBWebController(mainWindow);
     documentController = new UBDocumentController(mainWindow);
 
-    UBDrawingController::drawingController()->setStylusTool((int)UBStylusTool::Pen);
-
     applicationController = new UBApplicationController(boardController->controlView(),
                                                         boardController->displayView(),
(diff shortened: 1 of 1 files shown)


kaamui:
Thank you for the pull request. It actually works at start, but introduces a regression. After passing to Document Mode and then return to Board Mode, the stylus color is undefined. But this request helped me to find a solution that apparently work well. I'll commit it soon in the "dev" branch.

PR #196, opened 2019-02-03
Fix to build against later versions of OpenSSL

rogerjames99:
Quick fix for building against later versions of openssl. I am not familiar enough with OpenBoard to make this backward compatible with earlier versions of openssl. Maybe someone else can suggest a wrapper.


rogerjames99:
Withdrawing this to do tidy up on my repo.

PR #211, opened 2019-03-22
Port away from deprecated Qt APIs

JBBgameich:


JBBgameich:
status update: I rebased the changes and dropped the not directly related commits. I also ported away from APIs deprecated in newer Qt releases in a separate commit.

bbreinbauer:
Hi,
this breaks loading of documents for me. The following patch fixes it. (Background: The setMatrix() functions in QTransform and QMatrix have different parameters)
```
diff --git a/src/adaptors/UBSvgSubsetAdaptor.cpp b/src/adaptors/UBSvgSubsetAdaptor.cpp
index 6c613137..767ba575 100644
--- a/src/adaptors/UBSvgSubsetAdaptor.cpp
+++ b/src/adaptors/UBSvgSubsetAdaptor.cpp
@@ -122,11 +122,13 @@ QTransform UBSvgSubsetAdaptor::fromSvgTransform(const QString& transform)
         matrix.setMatrix(
                     sl.at(0).toFloat(),
                     sl.at(1).toFloat(),
+                    0,
                     sl.at(2).toFloat(),
                     sl.at(3).toFloat(),
+                    0,
                     sl.at(4).toFloat(),
                     sl.at(5).toFloat(),
-                    0, 0, 0);
+                    1);
     }
 
     return matrix;

```

JBBgameich:
Thank you for fixing it. This probably needs some more testing before it should be merged, since the diff is really large.

bbreinbauer:
I'm currently using a version with your patches (also the cmake patch, thank you very much for that!) and will report/fix any issues I find.

JBBgameich:
I removed the unrelated commits now, which unfortunately makes it not build for me anymore, but the branch should be mergeable here at least.

kaamui:
Hello,

Sorry but there are too much things that are not directly concerned by the topic of your pull request (to replace deprecated calls).

JBBgameich:
@kaamui Sorry about that, I didn't want it to be merged like this, but I based on a branch of mine so I could get it to compile while testing. If the commits that are actually about the deprecation porting are fine, I can drop the others.

JBBgameich:
I actually already had a branch prepared, so here is the clean version.

kaamui:
Some of these changes are located in very critical portions of the code, and with the schedule I have I'm not confident to be able to test it properly. I think I'm going to add this pull request right after the 1.6.1 is out, so the team can test it widely.

sorry for the time it takes, I hope you understand.

kaamui:
Also, what bug does cc5683c fix ?

JBBgameich:
> Also, what bug does [cc5683c](https://github.com/OpenBoard-org/OpenBoard/commit/cc5683ccd899d99c5d78844bb4a220c29948df1f) fix ?

It fixes a mistake I did in one of the earlier commits. See the discussion on this pull request for more details.
No problem if it takes longer. This will become really useful once you move to Qt 6, and I guess that will still take some time.

kaamui:
> I actually already had a branch prepared, so here is the clean version.

Hi @JBBgameich ,

I see in your fork that you have a `upstream-undeprecate-qt` branch that seems to only contain the work you did to remove deprecated calls, but I a bit confused by the quoted message. Did you forgot the link ? Did I misunderstood something ?

I would like to merge your work on the 1.7-dev branch as 1.6.2 is about to be released, but there's still some work not related to the subject in this PR.

JBBgameich:
Sorry about the mess in this merge request. I'll try to rebasr and clean this up later this week.

I think I got a bit confused with my own branches last time, but that should be easy to fix :)

JBBgameich:
The pull request should now have a manageable size.
I still consider it somewhat dangerous, due to the number of non-obvious ports from one API to another. With some testing, it should be fine I hope.

src/adaptors/UBMetadataDcSubsetAdaptor.cpp
```diff
@@ -31,6 +31,8 @@
 
 #include <QtGui>
 #include <QtXml>
+#include <QGuiApplication>
+#include <QScreen>
```
letsfindaway, line 35: These two added `#include`s are actually not needed, I think.

src/gui/UBKeyboardPalette.cpp
```diff
@@ -274,8 +274,8 @@ void  UBKeyboardPalette::paintEvent( QPaintEvent* event)
 
     QRect r = this->geometry();
 
-    int lleft, ltop, lright, lbottom;
-    getContentsMargins ( &lleft, &ltop, &lright, &lbottom ) ;
+    const auto margins = contentsMargins();
+    int lleft, ltop = (margins.left(), margins.top());
```
letsfindaway, line 278: I'm not sure what this line actually does on C/C++. But anyway, it would be sufficient to declare and initialize these two variables later in lines 284 and 285.

letsfindaway (changes requested):
Hi @JBBgameich,

I just shortly went through your changes, and for me they are all perfectly ok! I did not do extensive testing, but just some "code review". I just have two minor remarks, which, however, do not affect the operation or correctness of the program.

src/adaptors/UBMetadataDcSubsetAdaptor.cpp
JBBgameich: Oh right, the code that used them got lost during rebasing, as the original code was already replaced.

src/adaptors/UBCFFSubsetAdaptor.cpp
```diff
@@ -1206,7 +1206,9 @@ UBGraphicsGroupContainerItem *UBCFFSubsetAdaptor::UBCFFSubsetReader::parseIwbGro
 
 
 
-    foreach (QString key, strokesGroupsContainer.keys().toSet())
+    const auto keys = strokesGroupsContainer.keys();
+    const auto keySet = QSet<QString>(keys.begin(), keys.end());
+    for (const QString &key : keySet)
     {
```
letsfindaway, line 1212: Unfortunately, this leads to a compile error. And I think it is not necessary to create a `QSet`, you could as well iterate over the `QList`. My proposal would be:

```c++
    const auto keys = strokesGroupsContainer.keys();

    for (const QString &key : keys)
```

JBBgameich: It compiles correctly for me. Which Qt / gcc version are you compiling on?
Anyway, you are probably right about the set.

src/gui/UBKeyboardPalette.cpp
JBBgameich: Oh right! The initial value is not even used.

src/adaptors/UBCFFSubsetAdaptor.cpp
letsfindaway: Ok, I see. It needs Qt >= 5.14. The range constructor was added there. I'm compiling with 5.12 for some platforms. Could we find a solution which works for 5.12 to 6.2?

PR #266, opened 2020-01-20
Poppler >=0.83 

bartoszek:
Update `pdf renreder` to reflect latest poppler changes [poppler@gitlab](https://gitlab.freedesktop.org/poppler/poppler/commit/759d190581f8ff069ecee9155313a8e69a2ca9c6)
>    Thanks to loqs at
    https://bbs.archlinux.org/viewtopic.php?pid=1883212

### DIFF
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -3,6 +3,7 @@ TEMPLATE = app
 
 THIRD_PARTY_PATH=../OpenBoard-ThirdParty
 
+CONFIG += c++14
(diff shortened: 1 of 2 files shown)


kaamui:
Thanks a lot dude, you save me a lot of time, I'm so grateful of your work.

Btw, I just realized that the pre-processor directive `#if POPPLER_VERSION_MAJOR > 0` is useless now. I think it was relevant with xpdf, but I don't think it is necessary anymore.

PR #316, opened 2020-04-25
Handle HiDPI screens

multun:
Without this patch, OpenBoard looks like this with a HiDPI screen on wayland:
ko

With the patch, the scale is much more reasonable
ok

I'm targeting dev-qt5.1x as it's the branch that works for me

Can anyone confirm it doesn't break on windows? (I really don't expect it to, but who knows)


multun:
I think it was an xwayland issue? this patch causes other issues ever on linux

PR #320, opened 2020-04-26
Make Dev qt5.1x build on both windows and (Debian) Linux

ootwch:
Incorporated most (but not all!) changes from the debian package (built by @sunweaver) into the current branch. 

NOT done: All the java script libraries...

With these changes this package build on Debian Linux (once the dependencies are installed - taken from @sunweaver 's package). It does not require the Third Party Libs.

On Windows it links to Third Party libs and still needs them.

This works for me but has not been tested extensively - and I am _not_ an experienced CPP/QT developer.

### DIFF
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -1,8 +1,6 @@
 TARGET = "OpenBoard"
 TEMPLATE = app
 
-THIRD_PARTY_PATH=../OpenBoard-ThirdParty
-
 CONFIG += c++14
 CONFIG -= flat
 CONFIG += debug_and_release \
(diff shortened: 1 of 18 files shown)


kaamui:
Hello,

Thank you for your help !

Can you please make seperate PRs for each topic your commits refer to ?

For example,  I would like to pick what you did with XPDF, but I can't because of the other changes  I don't want at this time. 

To help you do it, in git, you should work in a topic-related branch to make PRs's commits only related to your request

to create a new branch from a previous commit (here the last commit in the dev-qt5.1x branch)

```
#git checkout -b <branch_name> <commit_sha1>
git checkout -b xpdf_win32 6edebf9632d977c60b5dcbf0ab63d2167b29e28a
```

Then you can cherry-pick the commits related to xpdf on windows, verify (at least) that it compiles and run. 

Thanks

PR #323, opened 2020-04-27
[DRAFT] draft live streaming for linux

multun:
This is not intended for production use, it's more like a proof of concept.
With just this code, I can plug OpenBoard with OBS and stream the board even when in the background!

It requires having v4l2loopback installed and some virtual device enabled.

Just run this after installing the package:
`sudo modprobe v4l2loopback devices=1`

`V4L2_DEVICE/dev/videoFIXME path/to/OpenBoard`


kaamui:
Hi @multun,

Thank you for this proof of concept. Don't know if there are still issues for OpenBoard to be detected in OBS, but closing for now.

PR #327, opened 2020-05-01
OpenBoard-in-a-Window

fsckawk:
With these changes normal compilation of OpenBoard won't change anything. But you can choose, when running "qmake", to create an OpenBoard executable which will run in its own window, instead of fullscreen.
I've tested in Linux, Windows and macOS.
http://webdiis.unizar.es/u/spd/openboard/

### DIFF
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -98,6 +98,11 @@ ALPHA_BETA_STR = $$find(VERSION, "[ab]")
 count(ALPHA_BETA_STR, 1):DEFINES += PRE_RELEASE
 BUILD_DIR = build
 
+OB_INAWINDOW = $$(OB_INAWINDOW)
+!isEmpty(OB_INAWINDOW) {
+	DEFINES += OB_INAWINDOW
+}
+
 macx:BUILD_DIR = $$BUILD_DIR/macx
 win32:BUILD_DIR = $$BUILD_DIR/win32
 linux-g++*:BUILD_DIR = $$BUILD_DIR/linux
@@ -432,6 +437,7 @@ linux-g++* {
     QMAKE_CFLAGS += -fopenmp
     QMAKE_CXXFLAGS += -fopenmp
     QMAKE_LFLAGS += -fopenmp
+	INCLUDEPATH += /usr/include/ffmpeg
     UB_LIBRARY.path = $$DESTDIR
     UB_I18N.path = $$DESTDIR/i18n
     UB_ETC.path = $$DESTDIR
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -437,7 +437,8 @@ linux-g++* {
     QMAKE_CFLAGS += -fopenmp
     QMAKE_CXXFLAGS += -fopenmp
     QMAKE_LFLAGS += -fopenmp
-	INCLUDEPATH += /usr/include/ffmpeg
+	# Necessary for CentOS/RHEL and won't harm in other distributions
+    INCLUDEPATH += /usr/include/ffmpeg
     UB_LIBRARY.path = $$DESTDIR
     UB_I18N.path = $$DESTDIR/i18n
     UB_ETC.path = $$DESTDIR
--- a/README-OB-IN-A-WINDOW.txt
+++ b/README-OB-IN-A-WINDOW.txt
@@ -1,0 +1,12 @@
+
+In order to compile a version of openboard which runs in it's own
+window instead of fullscreen, you should set the environment
+variable OB_INAWINDOW to yes before running qmake.
+
+For instance:
+env OB_INAWINDOW=yes qmake-qt5 OpenBoard.pro -spec linux-g++-64
+
+spd@daphne.cps.unizar.es
+https://webdiis.unizar.es/~spd/openboard
+
+
--- a/src/core/UBApplication.cpp
+++ b/src/core/UBApplication.cpp
@@ -279,7 +279,13 @@ int UBApplication::exec(const QString& pFileToImport)
 
 
 #ifdef OB_INAWINDOW
-    mainWindow = new UBMainWindow(0, Qt::WindowCloseButtonHint); // deleted by application destructor
+    mainWindow = new UBMainWindow(0,
+		Qt::Window |
+		Qt::WindowCloseButtonHint |
+		Qt::WindowMinimizeButtonHint |
+		Qt::WindowMaximizeButtonHint |
+		Qt::WindowShadeButtonHint
+	); // deleted by application destructor
 #else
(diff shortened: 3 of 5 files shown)


LawrenceRao007:
Thank you for this.  On Windows 10, it appears in a window when starting but no longer if the multi-screen option is enabled / disabled.

If it's easy to do, it would be very useful to be able to enable/disable the multi-screen option for 2 or more monitors.

fsckawk:
I will take a look. I've compiled the Windows version in a virtual machine, and I guess I could set it up with two (virtual) screens.

LawrenceRao007:
I think the problem is with 3 screens or more.

fsckawk:
I've tested with two virtual screens, and the problem is not that OB goes back to fullscreen mode. It seems the window frame is outside the limits of the main screen, but I was able to drag the left border of the window (from the right of the left -secondary- screen) and thus reduce the size of the window.
This seems to be related to another requested feature: on startup the window size should not be the same as the main screen, but a little bit smaller. I'd try to figure how to do that.

OpenBoard.pro
```diff
@@ -432,6 +437,7 @@ linux-g++* {
     QMAKE_CFLAGS += -fopenmp
     QMAKE_CXXFLAGS += -fopenmp
     QMAKE_LFLAGS += -fopenmp
+	INCLUDEPATH += /usr/include/ffmpeg
```
multun, line 440: probably unintended

fsckawk: No, that's intended. It's necessary for CentOS/RHEL and won't harm with other GNU/Linux distributions.

multun: I would add a comment explaining just that and indent with spaces then!
be careful, the commit name doesn't explain your change very well

fsckawk: OK, I've documented the change.
I've also noticed, that with Gnome3 ist necessary to modify src/frameworks/UBPlatformUtils_linux.cpp in the same way as src/frameworks/UBPlatformUtils_win.cpp

LawrenceRao007:
Hi, these changes are perfect - thank you.  I am using this software with my students every day during the lockdown and it has been invaluable.

There is another small fix that would be helpful - when you paste an image (CTRL+V) then the image is always pasted at the very top of the board rather than where I'm working.  I treat the work area as an A4 sheet of paper rather than a display board so I often work down below the 'display area' as the software sees it.  Being able to set the work area as well would also be helpful.

fsckawk:
> Hi, these changes are perfect - thank you. I am using this software with my students every day during the lockdown and it has been invaluable.
> 
> There is another small fix that would be helpful - when you paste an image (CTRL+V) then the image is always pasted at the very top of the board rather than where I'm working. I treat the work area as an A4 sheet of paper rather than a display board so I often work down below the 'display area' as the software sees it. Being able to set the work area as well would also be helpful.

I'll have a look.
Is this a problem with the original OpenBoard application too, or happens only in my modified version?

LawrenceRao007:
It was in the original.  I haven't got the skills to 'build' the new version that you've made - if that's the right word.

fsckawk:
OK, than the problem should be reported in the developers forum.
If you want to try my version, there are binaries in my page:
https://webdiis.unizar.es/u/spd/openboard/

LawrenceRao007:
Thanks, it still kills my second and third screens at startup.

I will report the paste bug in the developers forum.

yoanndaly:
Hi, thank you for this more convenient version of OpenBoard. In order to run the application downloaded [here](http://webdiis.unizar.es/u/spd/openboard/) on ubuntu 18.04, I needed to install first libquazip-dev_0.7.3-5ubuntu1_i386.deb Before, I had the error message 
`./OpenBoardW: error while loading shared libraries: libquazip.so.1: cannot open shared object file`

Thanks again.

x7i7l:
Hello, working on Ubuntu 20.04.
I followed your instructions for Ubuntu 18.04, but it wouldn't load
saying

> ./OpenBoardW: error while loading shared libraries: libquazip.so.1: cannot open shared object file: No such file or directory

So I installed libquazip5-dev from apt and still had the error message.

My solution was to manually copy "libquazip5.so.1.0.0" from usr/lib/x86_64-linux-gnu into the OpenBoardW's lib folder. I then renamed it "libquazip.so.1" and OpenBoard loads.

mosen85:
@fsckawk, thank you for your work!
Im missing the maximize-button on the right top-corner of the window. Is it possible to add this?

GavinSmith0123:
Thanks for this patch. I've had exactly the same issue with screen sharing software. With your patch, the program opened in a proper window.

It should be straightforward to add an option for this to the preferences. I may do this for my own builds of OpenBoard.

@fsckawk if I can ask, at your website http://webdiis.unizar.es/u/spd/openboard/ you say you offer for download a stand-alone Windows executable, which doesn't need installing. How did you produce this? Is using "windeployqt" as you say at http://webdiis.unizar.es/u/spd/openboard/compiling-windows.txt?

ggrinton:
Thank you for this helpful change. Having OpenBoard run in a Window is really essential in most environments.

I note that there is already a request above for minimise/maximise functionality. The current "Show Desktop" button does not, it seems to me, show the actual desktop, but rather an image of it. It is not possible to interact with the desktop icons or do anything else useful. A proper minimise function would make OpenBoard behave like every other application, getting it out of the way while the user wants to do other work. (Of course this request goes for the standard version as well, but this branch seems more likely to provide the functionality!)

PaulaUzca:
Thank you so much!!! My dad was having problems with his zoom classes because of this issue. He is very happy now thanks to your changes!

Remodor:
It is sooo unclear to me, that this is not an option. Great app, but a really poor design decision. This really complicates the life of teachers and co. I mean, common..

e-t-l:
> A proper minimise function would make OpenBoard behave like every other application, getting it out of the way while the user wants to do other work.

I've been using @fsckawk's OpenBoardW package for months and it's much more functional to have a resizable window. To follow up on @ggrinton's question, I just discovered Ctrl+H (on Windows) seems to minimize the application. Is it not possible to put this function (and a normal Maximize function) in the upper right on the status bar next to the close button, like normal app windows have?

RochiPacheco:
> With these changes normal compilation of OpenBoard won't change anything. But you can choose, when running "qmake", to create an OpenBoard executable which will run in its own window, instead of fullscreen. I've tested in Linux, Windows and macOS. http://webdiis.unizar.es/u/spd/openboard/

Thank you for this. As fas as I was using OB, It has been working fine with two screens.

kaamui:
A similar pull request has been merge in the 1.7-dev branch.

PR #334, opened 2020-05-03
Reviewed, reorganized and expanded the shape library

Gvf63:
- divided the shapes by type
- revised the code for all shapes eliminating redundancies and unnecessary data
- added green and yellow variants to existing shapes
- added the shapes: discs (solid color) and splash, both in 14 colors
- added emoticon
- added geometric shapes (plane geometry) also with letters and diagonals, heights etc
- added solid geometry shapes
- added Cuisenaire rods

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,4 +34,4 @@ Even with not a single bit of knowledge in C++, you can contribute to OpenBoard.
 
 Download and install OpenBoard, and develop your web app in it. You'll even find a web inspector to help you debug your site.
 
-You'll find documentation on how to create a Web Widget from scratch or tunring your already developped web site into an OpenBoard Web Widget [here](https://github.com/OpenBoard-org/OpenBoard/wiki/Creating-Web-Widgets).
+You'll find documentation on how to create a Web Widget from scratch or turning your already developped web site into an OpenBoard Web Widget [here](https://github.com/OpenBoard-org/OpenBoard/wiki/Creating-Web-Widgets).


Vekhir:
I not going to provide a code review, but I want to explain some issues with this PR that are making it difficult to integrate.
They all boil down to the fact that someone has to review the changes; it is not enough to just go "eh, it's probably fine".

1. The first thing to note is the number of files and changed lines. Obviously, smaller PRs get merged quicker. Though in this case it's reasonable due to the nature of the change.
2. The second comment concerns the commits. There are 5 commits of which 4 are, frankly, noise, so all the content is within one commit. I usually see commits as an opportunity to explain the changes and more importantly group them so each individual commit only does one thing (which might still involve several files, but similar changes). This makes the review process easier and allows it to be split up if necessary. While you could improve that, these commits would still be very large, or at least one would be.
3. This leads to the last point, and that's coincidentally also the thing you can probably best act upon, the fact that most changes are basically moves, except that you also minimised the files while you were at it. This means that each and every file has to be manually reviewed whether the changes affect anything or some bad behaviour has been introduced. GitHub itself detects moves and makes it easy to review that. But changing the content means that this cannot be done, making the review process incredibly tedious unnecessarily.

This PR tries to achieve too many things. In particular your second point just does too much at once. I cannot guarantee that it will be merged in another form, but I wouldn't expect this PR to be merged in its current form.

PR #336, opened 2020-05-05
fix use after free of movingItem

### LINKED ISSUE #259: Crash in UBBoardView::mouseReleaseEvent (Qt 5.13.1, with latest QtWebKit)
L.S.,

I've built my own QtWebKit (basically along the lines suggested in the #253 RoadMap issue. 

I can consistently crash the application in UBBoardView::mouseReleaseEvent near line 1244 (of the Qt5.1x branch), because movingItem is a dangling pointer:

```
UBGraphicsItem *graphicsItem = dynamic_cast<UBGraphicsItem*>(movingItem);
if (graphicsItem && graphicsItem->Delegate())
	graphicsItem->Delegate()->commitUndoStep();
```

1. Draw something on th eboard with the Pen tool
2. Use the selector tool, and select 1. by dragging a large rectangle around it
3. Click and release the mouse on 1. (inside the frame)

I am trying to figure out where the GraphicsItem is deleted that movingItem is still pointing to, but cannot find it at all. It also seems that the scene()->items() list is refreshed, i.e. the same number of items are in the list, but they all have a different address, between mouse clicks. Where is this happining?

I can implement an awful hack in Windows by enclosing the try ... catch(std::__non_rtti_object) block, but this will not work on Mac OSX.

Any ideas, pointers in the correct direction?

multun:
If an item marked as being moved is deleted, it then leaves a dangling pointer in the board view.
this bug caused a number of crashes for me. It can be reproduced easily by selecting an item, dragging it using its box, and deleting it using the delete key while not releasing the left click.

There are probably a number of other ways to reproduce it, I unfortunately did so way too many times before

fixes #259

### DIFF
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -95,6 +95,7 @@ UBBoardView::UBBoardView (UBBoardController* pController, QWidget* pParent, bool
     , mLongPressInterval(1000)
     , mIsDragInProgress(false)
     , mMultipleSelectionIsEnabled(false)
+    , _movingItem(nullptr)
     , bIsControl(isControl)
     , bIsDesktop(isDesktop)
 {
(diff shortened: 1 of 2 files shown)


kaamui:
hello,

thank you for that.It seems to work well, it fixes the issue you found and I did not found any regression. 

Please tell me if you find any.

PR #341, opened 2020-05-08
PR shortcuts

ootwch:
**Configurable Shortcuts**

- Replaced Event Handler with QShortcuts
- Added logic to "rotate" colors and width
- Fixed names of main window elements, so they are deterministic
- Added preference setting for the shortcuts
- Added "Advanced" Tab in preferences to edit the shortcuts (a bit raw - might need some love)

Known issues:
 - Upon startup no color is selected
 - Ctrl-I was used for two actions - "Zoom In" and "Pen". Now only "Pen"

**Note:** I tried to split up in single commits to better comment on what I did - the single commits will not all build, however.

Reference/Inspiration: https://doc.qt.io/archives/qq/qq14-actioneditor.html

### DIFF
--- a/resources/forms/mainWindow.ui
+++ b/resources/forms/mainWindow.ui
@@ -304,7 +304,7 @@
      <normaloff>:/images/toolbar/smallPen.png</normaloff>:/images/toolbar/smallPen.png</iconset>
    </property>
    <property name="text">
-    <string>Line</string>
+    <string>Small Line</string>
    </property>
    <property name="toolTip">
     <string>Small Line</string>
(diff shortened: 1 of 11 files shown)


kaamui:
hello,

Thank you for your reactivity. I'll test asap your work. 

> **Note:** I tried to split up in single commits to better comment on what I did - the single commits will not all build, however.

You should not do that. Every commit should be valid so the project can still compile. You can make one commit for all but comment your code to explain what you wanted to do. 

Thanks again !

kaamui:
I tested it, and it is a good start, but it seems to break some things. I cannot use the close button, and set a key to a color then use the "remove shortcut" button seem to remove the action.

It could be better also to have the action path displaying just the name of the action instead so it can be translated. I'll try to start from what you did to improve it. 

Thank you for your help !

PR #342, opened 2020-05-08
Improved (Wacom) Tablet handling

ootwch:
**Improved (Wacom) tablet handling.** 
Existing handling did not handle pressure well and added additional artefacts due to now actually working as described in the documentation.

- Fix handling of tablet events (Bug)
- Better (Prettier) handling of tablet (wacom) pressure (perhaps matter of taste)

**Reference:** https://stackoverflow.com/a/33525420/1878199

Note: This does not build on my Windows system without setting `USE_XPDF` - see my comment on the other commit.

```
#ifdef WIN32
    #define USE_XPDF
#endif
```

### DIFF
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -121,6 +121,7 @@ UBBoardView::UBBoardView (UBBoardController* pController, int pStartLayer, int p
     , mMultipleSelectionIsEnabled(false)
     , bIsControl(isControl)
     , bIsDesktop(isDesktop)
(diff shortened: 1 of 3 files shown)


kaamui:
Hi,

Closing your PR. The change is too "personal" as-is and we didn't have any complaint by our teachers. But I like agree with you that the 20% reduction comes from nowhere. Could it become an option in the settings, so the user can manage it himself ?

If you consider it and make it a PR, I'll add it to the next release. Please create a new PR, with only the code needed and corresponding to the topic of the PR

PR #343, opened 2020-05-08
Zoom with Ctrl-Mousewheel

ootwch:
Zoom with Ctrl-Mousewheel according to https://wiki.qt.io/Smooth_Zoom_In_QGraphicsView


andrewasd (approved):

SeeDborg:
I would add to this proposal : scroll the page up or down with the mouse wheel. It would be so useful !

Lucas-C:
I'd love to have this feature!

Thanks for submitting this PR @ootwch

letsfindaway:
To improve your PR I would like to add some suggestions:

- Please base your changes on `dev` and also request to merge to `dev`. The `dev-qt5.1x` branch is obsolete.
- Please make sure that your PR only contains the changes you want to make. The list of changed files https://github.com/OpenBoard-org/OpenBoard/pull/343/files contains many unrelated files.

I will also add some suggestions and questions to the code directly.

BTW: Finally I think a combined PR including scrolling with mouse wheel could be a good idea!

src/board/UBBoardView.h
```diff
@@ -99,6 +99,10 @@ class UBBoardView : public QGraphicsView
     virtual void mouseMoveEvent(QMouseEvent *event);
     virtual void mouseReleaseEvent(QMouseEvent *event);
     virtual void wheelEvent(QWheelEvent *event);
+
+    int _numScheduledScalings;
```
letsfindaway, line 103: Please use the naming conventions used in Openboard (starting a member variable with `m`) and place it where the other member variables are.

src/board/UBBoardView.cpp
```diff
@@ -112,6 +112,8 @@ UBBoardView::UBBoardView (UBBoardController* pController, QWidget* pParent, bool
 
     mLongPressTimer.setInterval(mLongPressInterval);
     mLongPressTimer.setSingleShot(true);
+
+    _numScheduledScalings = 0;
```
letsfindaway, line 116: Consider adding this to the initializer list.

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+        }
     }
+}
+
+void UBBoardView::wheelZoomEvent(QWheelEvent *event){
+    int numDegrees = event->delta() / 8;
+     int numSteps = numDegrees / 15; // see QWheelEvent documentation
+     _numScheduledScalings += numSteps;
+     if (_numScheduledScalings * numSteps < 0) // if user moved the wheel in another direction, we reset previously scheduled scalings
+     _numScheduledScalings = numSteps;
```
letsfindaway, line 1537: Use proper formatting (indentation), which makes code more readable.

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+
 
+
+     QTimeLine *anim = new QTimeLine(150, this);
```
letsfindaway, line 1541: Just a comment for proper understanding: This creates a new `QTimeLine` for each wheel event. Multiple timelines may run in parallel, if the next wheel event arrives before the previous timeline was finished.

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+     anim->setCurveShape(QTimeLine::LinearCurve);
+     anim->setUpdateInterval(10);
+
+     connect(anim, SIGNAL (valueChanged(qreal)), SLOT (wheelZoomEventScalingTime(qreal)));
+     connect(anim, SIGNAL (finished()), SLOT (wheelZoomEventAnimFinished()));
+     anim->start();
 }
+void UBBoardView::wheelZoomEventScalingTime(qreal x){
+    qreal factor = 1.0+ qreal(_numScheduledScalings) / 300.0;
```
letsfindaway, line 1550: When multiple timelines are running in parallel, this code increases the zoom probably beyond what was intended. First of all both timelines trigger this event, and then each of them uses the `_numScheduledScalings` to compute the factor.

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+
+    QPointF scenepoint =  this->mapToScene(QCursor::pos());
+
+    mController->zoom(factor, scenepoint);
+    // scale(factor, factor);
```
letsfindaway, line 1555: Remove commented code.

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+}
+void UBBoardView::wheelZoomEventAnimFinished(){
+    if (_numScheduledScalings > 0)
+    _numScheduledScalings--;
+    else
+    _numScheduledScalings++;
```
letsfindaway, line 1561: Use proper formatting (indentation).

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+    sender()->~QObject();
```
letsfindaway, line 1562: It is ugly (and in most cases not what you want to do) to directly invoke a destructor. This would destroy the object, but not release the associated memory. Consider using the `deleteLater` function.

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+                QGraphicsView::wheelEvent(wheelEvent);
+                wheelEvent->accept();
+            }
+
+        }
     }
+}
+
+void UBBoardView::wheelZoomEvent(QWheelEvent *event){
+    int numDegrees = event->delta() / 8;
```
letsfindaway, line 1533: The `delta` function is obsolete in Qt 5.15. Use one of the supported replacements.

src/board/UBBoardView.cpp
```diff
@@ -1492,29 +1494,74 @@ void UBBoardView::mouseDoubleClickEvent (QMouseEvent *event)
...
+     int numSteps = numDegrees / 15; // see QWheelEvent documentation
+     _numScheduledScalings += numSteps;
+     if (_numScheduledScalings * numSteps < 0) // if user moved the wheel in another direction, we reset previously scheduled scalings
+     _numScheduledScalings = numSteps;
+
 
+
+     QTimeLine *anim = new QTimeLine(150, this);
+     anim->setCurveShape(QTimeLine::LinearCurve);
+     anim->setUpdateInterval(10);
```
letsfindaway, line 1543: An update interval of 10ms is quite fast. In most cases the default of 40ms should be sufficient to create a smooth animation. Note that repainting may be expensive if you have many complicated objects on the screen.

letsfindaway (changes requested):
Please understand my comments as trying to be helpful, not as criticism. I think this is really a useful feature, and the smooth animation can improve user experience.

src/board/UBBoardView.cpp
letsfindaway, line 1537: Another remark on these lines: the code resets the `_numScheduledScalings` variable, but there might still be several `QTimeLine` objects running which fire their events. So if you e.g. create three wheel events in one direction and then turn back the wheel by one step, you will still end up with four `QTimeLine` instances creating their events mostly in parallel. With the current code it is not easily possible to cancel those. I wonder whether it would not be better to use a single `QTimeLine` instance and to adjust duration etc accordingly.

src/board/UBBoardView.cpp
letsfindaway, line 1550: Another remark on this line: Computing the scaling factor this way as `1 + x` causes that a zoom-in followed by a zoom-out does not end at the initial situation, but slightly offset. The error is not large, but it is inherent in this kind of formula. Just to make it clear with an example: if `x` is 0.1, then zoom-in uses a factor of `1 + 0.1 = 1.1`, and the zoom-out uses `1 - 0.1 = 0.9`. The sequential application of these factors yields `1.1 x 0.9 = 0.99` which is not 1.

letsfindaway (changes requested):
I added two more proposals for improvement.

kaamui:
a similar PR has been accepted

Issue #404, opened 2020-11-26
Import of SVG / SMART Notebook files


GavinSmith0123:
Is it possible to import arbitrary SVG files into OpenBoard, and if not, what are the limitations on this?

I am using the package for Ubuntu `openboard_ubuntu_18.04_1.5.3_amd64.deb` (the earlier, official release didn't work for me).

Import of SVG files would be greatly useful for migrating SMART Notebook files.

The attached file contains an SVG of a file I extracted from a SMART Notebook file [testpage.zip](https://github.com/OpenBoard-org/OpenBoard/files/5603040/testpage.zip).  OpenBoard appears to use SVG internally but I couldn't get the page to display by copying this SVG into one of the directories for OpenBoard and renaming it as something like `page001.svg`.  `openboard` outputs an error: `error parsing file  "Premature end of document."` when I try to open that page.

If this can be made to work, OpenBoard seems to be the most promising option for migrating away from SMART Notebook for an educational organization. I believe many organizations are looking to abandon SMART Notebook due to their move to a subscription-based business model (I'm not privy to all the details here). However, other options have imperfect support of importing SMART Notebook files: text formatting is often messed up. This seems unnecessary, as when you open the Notebook file as a .zip and look at the .svg pages individually, they display perfectly. SMART Notebook software itself can export to a `*.iwb` format, which also messes up text formatting.

The issue with text formatting appears to relate to the handling of the SVG `<text>` element (the page I attached uses such an element), and use of `x` and `y` attributes on contained `<tspan>` elements. This is not supported by the `*.iwb` format [spec is linked here](https://www.imsglobal.org/iwbcff/index.html) and SMART Notebook's export uses the `<textarea>` element instead.

So if OpenBoard had full support for `<text>` and `<tspan>`, that would be a big start.

SMART Notebook can also embed Flash objects, but these could be dropped when exporting is done.

GavinSmith0123:
After downloading and building the sources from `master`, I tried to find what is happening with the sample svg I posted.

A `<text>` element is processed by `loadScene` in `src/adaptors/UBSvgSubsetAdaptor.cpp`, calling `textItemFromSvg`. This appears to be the wrong function, though, as it doesn't handle any `<tspan>` elements. (`<tspan>` are dealt with elsewhere in the code, in `adaptors/UBCFFSubsetAdaptor.cpp`). `textItemFromSvg` chews through the rest of the file, leading to a

```
error parsing file  "Premature end of document."
```

[...]

PR #414, opened 2020-12-20
SMART Notebook import

GavinSmith0123:
Linked to issue #404 .

Here is the work I have done on creating a SMART Notebook importer. It is not complete, but I think it is the minimum that would be useful. Importing text, images and pen strokes is supported.

Before I continue work on missing features (like page backgrounds and supporting shapes like rectangles and circles), it would be very helpful to have some feedback before the code diverges too much from the original, and to avoid creating an incompatible version of OpenBoard.


GavinSmith0123:
I did a `git merge` instead of `git rebase`, now it seems I'm in git hell...

GavinSmith0123:
Replaced with fresh pull request with fixed git history. https://github.com/OpenBoard-org/OpenBoard/pull/424

PR #424, opened 2020-12-30
SMART Notebook import

GavinSmith0123:
Linked to issue #404 .

Here is the work I have done on creating a SMART Notebook importer. It is not complete, but I think it is the minimum that would be useful. Importing text, images, pen strokes and page backgrounds is supported.

Before I continue work on missing features (like supporting shapes like rectangles and circles), it would be very helpful to have some feedback before the code diverges too much from the original, and to avoid creating an incompatible version of OpenBoard.

### DIFF
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -438,7 +438,7 @@ linux-g++* {
     #LIBS += -lprofiler
     LIBS += -lX11
     LIBS += -lquazip5
-    INCLUDEPATH += "/usr/include/quazip"
+    INCLUDEPATH += "/usr/include/quazip5"
 
     LIBS += -lpoppler
     INCLUDEPATH += "/usr/include/poppler"
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -443,9 +443,9 @@ linux-g++* {
     LIBS += -lpoppler
     INCLUDEPATH += "/usr/include/poppler"
 
-    QMAKE_CFLAGS += -fopenmp
-    QMAKE_CXXFLAGS += -fopenmp
-    QMAKE_LFLAGS += -fopenmp
+    QMAKE_CFLAGS += -fopenmp -pg
+    QMAKE_CXXFLAGS += -fopenmp -pg
+    QMAKE_LFLAGS += -fopenmp -pg
     UB_LIBRARY.path = $$DESTDIR
     UB_I18N.path = $$DESTDIR/i18n
     UB_ETC.path = $$DESTDIR
--- a/src/adaptors/UBImportSMART.cpp
+++ b/src/adaptors/UBImportSMART.cpp
@@ -1,0 +1,323 @@
+/*
+ * Copyright (C) 2015-2018 Département de l'Instruction Publique (DIP-SEM)
+ *
+ * Copyright (C) 2013 Open Education Foundation
+ *
+ * Copyright (C) 2010-2013 Groupement d'Intérêt Public pour
+ * l'Education Numérique en Afrique (GIP ENA)
+ *
+ * This file is part of OpenBoard.
+ *
+ * OpenBoard is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, version 3 of the License,
+ * with a specific linking exception for the OpenSSL project's
+ * "OpenSSL" library (or with modified versions of it that use the
+ * same license as the "OpenSSL" library).
+ *
+ * OpenBoard is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with OpenBoard. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+
+#include <QDir>
+#include <QList>
+#include <QXmlStreamReader>
+
+#include "adaptors/UBSvgSubsetAdaptor.h"
+#include "core/UBApplication.h"
+#include "core/UBPersistenceManager.h"
+#include "core/UBDocumentManager.h"
+#include "core/UBPersistenceManager.h"
+#include "document/UBDocumentProxy.h"
+#include "frameworks/UBFileSystemUtils.h"
+
+#include "UBSvgSubsetAdaptor.h"
+
+#include "UBImportSMART.h"
+
+#include "globals/UBGlobals.h"
+
+THIRD_PARTY_WARNINGS_DISABLE
+#include "quazip.h"
+#include "quazipfile.h"
+#include "quazipfileinfo.h"
+THIRD_PARTY_WARNINGS_ENABLE
+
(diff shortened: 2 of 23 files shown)


OpenBoard.pro
```diff
@@ -438,14 +438,14 @@ linux-g++* {
     #LIBS += -lprofiler
     LIBS += -lX11
     LIBS += -lquazip5
-    INCLUDEPATH += "/usr/include/quazip"
+    INCLUDEPATH += "/usr/include/quazip5"
```
JBBgameich, line 441: (Disclaimer, I'm not a maintainer)

This is distribution specific, an actual fix needs to check multiple paths + check pkgconfig. I'd be in favor of using CMake for the project, since its better suited for finding external dependencies, but this is not plannend upstream as far as I know.

src/adaptors/UBImportSMART.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+
+
+QString UBImportSMART::importFileFilter()
+{
+    QString filter = tr("SMART Notebook (");
+    QStringList formats = supportedExtentions();
+    bool isFirst = true;
+
+    foreach(QString format, formats)
```
JBBgameich, line 81: foreach is old Qt api, I'd suggest to make `formats´ const, and iterate through it using
```
for (const QString &format : formats) {
}
```

src/adaptors/UBImportSMART.cpp
```diff
@@ -0,0 +1,346 @@
...
+    qreal adjust_x = scene->width() / 2;
+    qreal adjust_y = scene->height() / 2;
+
+    // Do not use the original height of the page, as the
+    // page is scaled to appear on the screen.  Use a 4:3 ratio
+    // instead so the page is large enough to read.
+    qreal newHeight = scene->width() * 0.75;
+    adjust_y -= 0.5 * (scene->height() - newHeight);
+
+    for (QGraphicsItem* item : scene->items())
```
JBBgameich, line 211: Iterating on a non-const container can cause detaches,
```suggestion
    const auto sceneItems = scene->items();
    for (QGraphicsItem* item : sceneItems)
```

src/adaptors/UBSvgSubsetAdaptor.cpp
```diff
@@ -1448,6 +1700,31 @@ bool UBSvgSubsetAdaptor::UBSvgSubsetWriter::persistScene(UBDocumentProxy* proxy,
     return true;
 }
 
+void
+UBSvgSubsetAdaptor::UBSvgSubsetWriter::persistDefs()
+{
+  QPixmap bgPattern = mScene->bgPattern();
```
JBBgameich, line 1706: Indentation seems off here.

src/adaptors/UBSvgSubsetAdaptor.cpp
```diff
@@ -2606,6 +2914,271 @@ void UBSvgSubsetAdaptor::UBSvgSubsetWriter::textItemToSvg(UBGraphicsTextItem* it
...
+{
+#if 0
+  qreal dxr = 0.0;
+  qreal dyr = 0.0;
+  qreal dx = 0.0;
+  qreal dy = 0.0;
+  qreal angle = 0.0;
+#endif
+
+  foreach(QString trStr, trString.split(" ", QString::SkipEmptyParts))
```
JBBgameich, line 3151:
```suggestion
  const auto splitTrString = trString.split(" ", QString::SkipEmptyParts);
  for (const QString &trStr : splitTrString)
```

src/adaptors/UBSvgSubsetAdaptor.cpp
```diff
@@ -2606,6 +2914,271 @@ void UBSvgSubsetAdaptor::UBSvgSubsetWriter::textItemToSvg(UBGraphicsTextItem* it
...
+  {
+      QRegExp regexp;
+#if 0
```
JBBgameich, line 3154: dead code

JBBgameich (changes requested):
This seems very useful. I am not a maintainer of this project, but I hope my suggestions can be of some use anyway :)

src/adaptors/UBImportSMART.cpp
```diff
@@ -0,0 +1,346 @@
...
+{
+    QStringList formats = {"notebook", "xbk"};
+    return formats;
+}
+
+
+QString UBImportSMART::importFileFilter()
+{
+    QString filter = tr("SMART Notebook (");
+    QStringList formats = supportedExtentions();
```
JBBgameich, line 78:
```suggestion
    const QStringList formats = supportedExtentions();
```

src/adaptors/UBSvgSubsetAdaptor.cpp
GavinSmith0123: It's there in case those parts of the transformation will be supported in the future. I'll remove the code if the maintainers ask me to.

OpenBoard.pro
GavinSmith0123: Thanks, do you know if there is a way for me to have that patch in my repository so I can build the program myself but not included in the pull request?

GavinSmith0123:
Thanks for your review @JBBgameich ; I've applied most of your suggestions.

The one part of the code I found confusing was the "strokes group". Did you look at or understand the changes I made?

It appears that as part of "stroke optimization", multiple polygon items can be output for a single stroke. However, when they are persisted, they are combined into a single SVG element. When read in again there will just be one polygon item in the graphics scene. I suspect the whole "strokes group" concept could be abandoned, but this is making even more changes.

This interacts with element grouping which is not done correctly at the moment for SMART import.

Did you try my code with any SMART Notebook files?

src/adaptors/UBSvgSubsetAdaptor.cpp
GavinSmith0123: I should add I copied and pasted this code from the CFF Import code (which doesn't work, as far as I can tell).

OpenBoard.pro
JBBgameich: I'm facing this problem myself, since I have lots of local patches I need for OpenBoard to properly work on my "too new" system. So far I have worked in a branch containing those patches, and use `git rebase -i origin/${BRANCH}` in the end to throw out the unrelated changes before merging.

JBBgameich:
Sorry, I don't have any SMART Notebook files to test with right now. I'm just trying to reduce the review load for the maintainers a bit by doing the generic Qt review part, since the maintainers usually have very limited amount of time.

OpenBoard.pro
GavinSmith0123: I attempted this and github detected that I'd deleted the branch while I was trying to move it, and closed this pull request. So I've made a fresh pull request for the new branch. https://github.com/OpenBoard-org/OpenBoard/pull/433/files (Unfortunately there are still extraneous changes there too, but I don't want to have to create yet another pull request to fix it.)

What I did in my repository is to add a branch "build-fixes" with the extraneous changes on it, based on the "dev-qt5.1x" branch. I then used git cherry-pick to get rest of the changes, and made this branch "smart-import-build". Finally, I used "git cherry-pick build-fixes..smart-import-build" on a new "smart-import" branch to extract the commits I wanted in the pull request.

PR #433, opened 2021-01-17
SMART Notebook import

GavinSmith0123:
Linked to issue #404. Recreated pull request with some commits removed. (Old pull request was https://github.com/OpenBoard-org/OpenBoard/pull/424).

Here is the work I have done on creating a SMART Notebook importer. Importing text, images, pen strokes, shapes and page backgrounds is supported.

Before I continue work on missing features it would be very helpful to have some feedback before the code diverges too much from the original, and to avoid creating an incompatible version of OpenBoard.

### DIFF
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -443,9 +443,9 @@ linux-g++* {
     LIBS += -lpoppler
     INCLUDEPATH += "/usr/include/poppler"
 
-    QMAKE_CFLAGS += -fopenmp
-    QMAKE_CXXFLAGS += -fopenmp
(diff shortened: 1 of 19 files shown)


kaamui:
Hi,

too much changes to add the support of something that is not asked by our teachers. Added the point multiple times through the years internally but it is not something known or used here. 

What I can add is that we have similar feature requests, like being able to import PowerPoint or Word files, but we don't want to go there and want OpenBoard to remain simple. So I must decline this PR, despite the amount of effort put in it. You can still create a fork and package a version with the support of it if you really need to and want to share this support with others.

Issue #445, opened 2021-02-21
Feature: draw cartesian coordinate system and intermediate grid lines


letsfindaway:
In practical use my wife was missing a tool to easily draw a cartesian coordinate system as it is often used in geometry. As at least in Germany the pupils are typically using a squared paper with 5mm grid lines, she also asked me whether it would be possible to draw intermediate grid lines, so that 1 unit corresponds to two grid lines.

I then implemented these features and will shortly also issue a merge request. The additional features are:

- a built-in tool which draws a cartesian coordinate system called "axes".
- the axes can be dragged around and will always snap to the grid
- numbering can be switched on and off
- the axes can be resized at all four arrow ends
- for small grid sizes where all numbers won't fit, only the 1 and all multiples of 5 are shown
- the background palette got an additional checkbox to switch on/off drawing of intermediate grid lines

If you feel these features are useful for OpenBoard I would be happy if they could be merged. More details will follow in the merge request.

My build for openSUSE is available at https://build.opensuse.org/package/show/home:letsfindaway/OpenBoard and can be installed for Leap 15.2 and Tumbleweed from https://software.opensuse.org/package/OpenBoard as community package.

kaamui:
Hello,

Thank you for (all !^^) your feedbacks. 

I'll try to look at it this month, but the code looks good. I didn't look at all so it might be off topic but did you consider making it as an html application, or was it impossible considering what you wanted to achieve ?

OpenBoard provides the ability to drag and drop html applications (and you can create your owns and add them to you OpenBoard app) directly on a page of an OpenBoard document.

letsfindaway:
Hi,

yes, I also checked the possibility to do this as an HTML application. As far as I can see, these applications just get a rectangular canvas and can do there whatever they want using JavaScript. They do not have any further integration into the core and cannot interact with the core. They are more or less "sandboxed" and probably do not even know their position on the scene.

What the axes application needed was:

- Know the background grid size to adapt automatically.
- Know the background grid line positions to snap-to-grid.
- Be non-rectangular. The active area of this application is some cross-shaped range just around the axes.
- Have a fixed zero-point. Resizing to the left should resize the negative x axis, resizing to the right should resize the positive x-axis and so on. When resizing an HTML application just the canvas size changes.
- Adapt the drawing color to the background color.

All these have been reasons to have this integrated in the core.

[...]

PR #446, opened 2021-02-21
Tool axes+intermediate lines

letsfindaway:
This pull request implements the features proposed in #445.

It was developed on top of dev-qt5.1x, but could also be merged on top of master.

- The axes tool was developed using the existing tools as blueprint. It also implements persistence.
- The intermediate grid lines are an extension to the background palette. They work for lined and for cross backgrounds. The setting is also persisted.
- The tools require three additional translation strings. I have only added a German translation, as this is my language.

The merge request consists of three commits for axes, grid lines and translations. They can also be applied individually, as they do not build on each other.

If any changes are required before merging, I will be happy to supply them.


letsfindaway:
The latest commit now disables the new features by default, so that users re not disturbed by an undocumented feature. The axes tool and intermediate lines can be enabled by the following settings:

```
[Board]
EnableIntermediateLines=true
EnableToolAxes=true
```

The features are merely hidden from the user interface, they are not completely disabled. So when a document containing axes and/or intermediate lines is created on an instance of OpenBoard with these features enabled and then opened on another (or the same) instance with these features disabled, then they are still being displayed.

Another change made by the latest commit is, that the axes are now also part of a screen capture or a PDF export.

Issue #340, opened 2020-05-07
Possibility to run on a separate window 


vitsca:
HI. I would like the openboard to be run on a separate (i.e. resizable) window. I have read the forum thread at  https://openboard.ch/forum/viewtopic.php?t=931 and I think I disagree with the arguments there. I understand the "default" full screen mechanism (for unexperienced users), but an experienced user should be able to have one (or both windows) to be windowed. 
Nowadays, with the COVID19 emergency it is even more important: on the "control" screen with the online teaching we all do, it is important to have the control of the class (via Zoom, Microsoft Teams, Meet or others) AND the editing of the board, while the other screen is used to show slides, the whiteboard etc. 
It would really make a hit if, at least, the editing window could be resizable. Thanks a lot!

profejuanca:
I read the post, windowed is a difficult thing? no sense in this point I think, every program is windowed and every person assumes this as a fact.

profejuanca:
Maybe for interactive whiteboards this approach is the best because Openboard don't coexist with other software at the same time or don't need to be streamed over Internet, but in many situations in the actual context window option is very important.

kaamui:
Hello,

As I already said, if the option is configurable, disabled by default, and editable via a checkbox in the display tab of the preferences, I'm sure it will not bother anyone. 

@fsckawk already made a PR. I think He found an issue with multi-screen mode, though.  Just clean it up, add the "default disabled" configuration option, add logic to be able to modify it from the preferences, and I'll add it. 

What we don't want is to make changes that force our teachers to relearn how to use OpenBoard at every patch. So all you need is to make it transparent for standard users, but available for the advanced ones.

There's a ton of examples in the commits history where I add a configurable option, it's quite easy. 

The only difficulty may reside in making this feature behave properly in multi-screen (go back to full screen (and disable the option if multi-screen activated)), or make it work with multi-screen (but not sure we want that for secondary screens), or maybe maintain main screen windowed but others full-screen (but again, not sure if it a useful use case).

vitsca:
> What we don't want is to make changes that force our teachers to relearn how to use OpenBoard at every patch. So all you need is to make it transparent for standard users, but available for the advanced ones.

[...]

PR #497, opened 2021-06-20
Adds RunInWindow flag to OpenBoard

micolous:
This patch is based on work by @fsckawk and @sagredo-dev in #327 and #340:

* removed the "RunInWindow patch" comments, because this is no longer "just a patch" :)
* removed additional ffmpeg include (which appears to be another bug, not related to this issue)
* fix indentation

My intent in opening this PR is to make sure the patch is not lost.  It still requires editing the configuration file, but this is significantly easier than building OpenBoard from source.

I think that will help OpenBoard users:

* use it with screen sharing software for remote learning
* use it with OBS or other video mixing software
* use it on computers with more than 2 monitors (if it's a window, it can be put anywhere)

There are [community OpenBoard builds with this or similar functionality](https://webdiis.unizar.es/~spd/openboard/), based on old versions of OpenBoard which may have other issues. There are also many forum posts about this issue since 2017.

**Note:** I'm unable to compile or test this properly because [Qt have resumed restricting downloads to those with a Qt account as of 2020](https://www.qt.io/blog/qt-offering-changes-2020), [despite a previous walk-back in 2015](https://www.qt.io/blog/2015/05/06/changing-qt-account-to-be-optional-in-the-online-installer). This also makes it impossible for me to work on a GUI for this option. 😞

### DIFF
--- a/resources/etc/OpenBoard.config
+++ b/resources/etc/OpenBoard.config
@@ -14,6 +14,7 @@ OnlineUserName=
 PageCacheSize=20
 PreferredLanguage=fr_CH
 ProductWebAddress=http://www.openboard.ch
+RunInWindow=false
 SoftwareUpdateURL=http://www.openboard.ch/update.json
 StartMode=
 SwapControlAndDisplayScreens=false
--- a/src/core/UBApplication.cpp
+++ b/src/core/UBApplication.cpp
@@ -276,8 +276,18 @@ int UBApplication::exec(const QString& pFileToImport)
     gs->setAttribute(QWebSettings::JavascriptCanAccessClipboard, true);
     gs->setAttribute(QWebSettings::DnsPrefetchEnabled, true);
 
+    if (UBSettings::settings()->appRunInWindow->get().toBool()) {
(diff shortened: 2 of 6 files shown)


letsfindaway:
*Note:* Qt5 up to QT 5.15 is still available freely and without any account. The change you mentioned only affects any versions from Qt 5.16 upwards. All Linux distributions offer devel packages for Qt 5. Also the links provided in the Wiki (https://github.com/OpenBoard-org/OpenBoard/wiki/Build-OpenBoard-on-Windows-10) still work. So there is nothing which prevents you from compiling and testing OpenBoard ;-)

kaamui:
showNormal should not be used as it restores the standard ratio of the window : 
https://doc.qt.io/qt-5/qwidget.html#showNormal

[show()](https://doc.qt.io/qt-5/qwidget.html#showNormal) would be better I think. 

Were you able to test the behavior of windowed mode while in a multiple screens context ?

I see no changes in UBPlatformUtils_mac.mm. Is it intended (better behavior as is ?)

PR #500, opened 2021-07-10
Fix support hand shortcut (ctrl + h)

bahmannejati:
Hi everyone,

There is no shortcut for the Hand tool (known as `Scroll Page` in stylus tools), so tried to add this shortcut. It's very annoying to click the button without any shortcut! I guess it's useful.

I don't know anything about C++ and didn't build the project to see if this works, can anyone check this?
It seems be alright when I read the code, hope it works :)

### DIFF
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -288,6 +288,12 @@ void UBBoardView::keyPressEvent (QKeyEvent *event)
                 event->accept ();
                 break;
             }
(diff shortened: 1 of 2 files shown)


kaamui:
Hi,

Actually H is already in used in the application to hide OpenBoard. But I don't like it and prefer your proposal. I cannot accept your PR as-is, but will introduce your proposal when I add the possibility to move the board by holding the middle button of the mouse.

Issue #491, opened 2021-05-31
[gcc11] definition collision in pdf-merger


bartoszek:
As of `gcc:11` there's a `void _serialize()` macro defined for the Intel ISA for an architecture specific serialization https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=366386c7a970bf6a32c5a2abb714ca6c7f4e7790

This seems to introduce a collision with `Object::_serialize()` defined in `pdf-merger/Object.h`

https://github.com/OpenBoard-org/OpenBoard/blob/f3f07f9930660a29eaa3cc95c86d6534109c8931/src/pdf-merger/Object.h#L155

As a rule of thumb one should omit using `under` or `dubleunder` prefix when naming entities in `c/cpp` code to avoid collisions with compiler specific names :feelsgood: feel free to rename it to whatever you like, but as a general solution would suggest renaming all variables and functions staring with the `_`

`gcc:11.log`
```
In file included from src/pdf-merger/Page.h:34,
                 from src/pdf-merger/Document.h:33,
                 from src/pdf-merger/Merger.h:33,
                 from src/adaptors/UBExportFullPDF.cpp:52:
src/pdf-merger/Object.h:155:72: error: macro "_serialize" passed 2 arguments, but takes just 0
  155 |        void _serialize(std::ofstream  & out, const std::string & stream);
      |                                                                        ^
In file included from /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/include/x86gprintrin.h:71,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/include/immintrin.h:27,
                 from /usr/include/qt/QtCore/qfloat16.h:59,
                 from /usr/include/qt/QtCore/qcborstreamreader.h:45,
                 from /usr/include/qt/QtCore/qcborstream.h:47,
                 from /usr/include/qt/QtCore/QtCore:41,
                 from src/adaptors/UBExportFullPDF.h:33,
                 from src/adaptors/UBExportFullPDF.cpp:30:
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/include/serializeintrin.h:37: note: macro "_serialize" defined here
   37 | #define _serialize()    __builtin_ia32_serialize ()
      |
In file included from src/pdf-merger/Page.h:34,
                 from src/pdf-merger/Document.h:33,
                 from src/pdf-merger/Merger.h:33,
[...]

PR #528, opened 2021-10-09
refactor: remove leading underline from all symbols in pdf-merger

letsfindaway:
This pull request would resolve issue #491

Even if underscores as leading character are legal in some context, I refactored to avoid any such symbols:

- global symbols
- members
- fields

by

- remove the leading underscore of all functions
- append "Impl" to private functions when necessary
- replace the leading underscore of fields by "m_"
- replace the leading underscore of static fields by "s_"

Additionally:

- remove pdf-merger/PageParser.h as it is never referenced


letsfindaway:
I set this to WIP because I wil provide another PR later which only refactors linker symbols starting with underline, leading to much fewer changes.

letsfindaway:
Anyway the changes of this PR are much too many. Closing this PR as not being really needed.

PR #545, opened 2021-12-04
Webengine improvements

letsfindaway:
This PR bundles improvements to the already merged work on replacing QWebKit by QWebEngine. It will remain in Draft status as long as needed to collect further improvements or issues detected during testing.

I encourage all to test this version. Issues related to web browsing may be directly reported on my fort at https://github.com/letsfindaway/OpenBoard/issues.

Currently this PR contains the following improvements:

- fix: cleanup scenes from cache. This deletes `QWebEnginePages` at shutdown to allow a clean shutdown of the `QWebEngine` without pending users of `QWebEngineProfiles`
- refactor: remove unused member variable from `BrowserWindow`. Just what it says.
- use default `QWebEngineProfile`. Instead of using a named profile we're now using the default profile. This enables using the option `--single-process` when starting OpenBoard, which helps debugging issues related to `QWebEngine`.
- feature: cookie handling and private browsing. This synchronizes cookies between `QWebEngine` and `QNetworkAccessManager` and adds cookie policies and an auto-delete on exit option.
- fix: Full save of browser history when items have expired.


letsfindaway:
@kaamui : I could also provide the commit 7713cdc as a separate PR to `dev`, as it fixes a bug which is independent of `QWebEngine` and was already there before my port. Of course you can also commit this simple one-line-fix by yourself if you like.

src/core/UBSceneCache.cpp
```diff
@@ -49,7 +49,8 @@ UBSceneCache::UBSceneCache()
 
 UBSceneCache::~UBSceneCache()
 {
-    // NOOP
+    // delete all cached scenes
```
letsfindaway, line 52: This provokes a sporadic crash during shutdown. To be analyzed under which circumstances a scene would be deleted twice.

letsfindaway: Fixed. Removed the change on the Scene Cache, which was probably not the right thing to do. Changing ownership of web views was sufficient.

kaamui:
Hi @letsfindaway,

can you make a separate PR for the "scaling of mirror pixmap" ?

thx

letsfindaway:
> Hi @letsfindaway,
> 
> can you make a separate PR for the "scaling of mirror pixmap" ?
> 
> thx

Of course! Will do this in the next days.

letsfindaway:
> > Hi @letsfindaway,
> > can you make a separate PR for the "scaling of mirror pixmap" ?
> > thx
> 
> Of course! Will do this in the next days.

Done. Also removed that commit from this PR.

PR #551, opened 2022-01-05
refactor: drawBackground mostly in scene

letsfindaway:
This PR harmonized the drawing of the background in the scene and the view, which is currently duplicated with mostly identical code. It does:

- move all drawing of background grid to UBGraphicsScene
- only add document border in UBBoardView
- avoid some compiler warnings

This PR should be merged before enabling the intermediate background grid lines by default, as currently the intermediate lines are not drawn in the magnifier tool due to an inconsistency in the two implementations.

See also discussion at https://github.com/letsfindaway/OpenBoard/issues/12

### DIFF
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -29,6 +29,7 @@
 
 #include "UBBoardView.h"
 
+#include <QtGlobal>
 #include <QtGui>
 #include <QtXml>
 #include <QListView>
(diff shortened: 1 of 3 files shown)


letsfindaway:
This PR now also fixes #710 and avoids a set of clang warnings for `UBGraphicsScene`.

letsfindaway:
Resolved conflicts. Note that this PR also contains a fix for #710, which conflicted with the merged fix. I think the solution of this PR is more complete and cleaner.

PR #572, opened 2022-03-11
Fix compile error in UBPlatformUtils on 1.7-dev branch

JBBgameich:
The static variable `errorOpeningVirtualKeyboard` was previously only declared on OSX, but referenced unconditionally.


letsfindaway:
This was also fixed with 9169146e272023cad7e5accdd555de7d243d1d4f on the `dev` branch, which is however still not merged to `1.7-dev`.

PR #582, opened 2022-04-02
feat: enable hidden features

letsfindaway:
Since more than a year there are two "hidden" features in OpenBoard:

- A tool to draw a Cartesian coordinate system.
- An option to draw intermediate grid lines to resemble the look-and-feel of typical math exercise books.

I think it is now time to offer these tools to all users. This PR provides the following:

- enable the axes tool unconditionally
- enable option for intermediate grid lines unconditionally
- remove associated settings

See also the merged PR #446


kaamui:
Hi @letsfindaway,

the QA team could finally review the Axes tool and has some feedback to give: 
- arrows in negative coordinates should not be displayed, as it is more common to draw axes with just arrows in the positive range
- when reducing the size of the left side of the x axis, rendering is not perfectly updated and some little artiacts can remain
- the axes tool behaves like the ruler, triangle, etc, but it makes it difficult to use : when exporting a document, and importing in another machine with a different configuration (gridSize, resolution, etc), these config changes alter the distance between two points of the axes and make the drawing made initially above these points not aligned anymore. It should preserve the initial size it had when dropped to avoid the drawing not corresponding anymore.

The last point is imo an important one. I can't see a good way to concile both behaviors, and I see the advantage of the current one, but the fact that the axes and a drawing made above could no longer correspond is an unwanted behavior. If you have a solution to maintain both, I would be happy to hear it. If not, it should be changed to be more consistent across machines. 

For the little graphical glitches, I could reproduce it (a tiny bit of the minus sign remains visible, not systematic) but it's not a mandatory thing to fix to make tool axes enabled by default in the next release. 

With these changes made, the tool axes (and the intermediate grid line too) will be enabled by default, and available as-is in the 1.7 :)

letsfindaway:
Hi @kaamui,
> Hi @letsfindaway,
> 
> the QA team could finally review the Axes tool and has some feedback to give:
> 
> * arrows in negative coordinates should not be displayed, as it is more common to draw axes with just arrows in the positive range

agreed, will change that.
> 
> * when reducing the size of the left side of the x axis, rendering is not perfectly updated and some little artiacts can remain
will look into this, too. I think I also have seen that. Rendering could probably improve.
> 
> * the axes tool behaves like the ruler, triangle, etc, but it makes it difficult to use : when exporting a document, then importing it in another machine with a different configuration (gridSize, resolution, etc), these config changes alter the distance between two points of the axes and make the drawing made initially above these points not aligned anymore. It should preserve the initial size it had when dropped to avoid the drawing not corresponding anymore.
> 
> The last point is imo an important one. I can't see a good way to concile both behaviors, and I see the advantage of the current one, but the fact that the axes and a drawing made above could no longer correspond is an unwanted behavior. If you have a solution to maintain both, I would be happy to hear it. If not, it should be changed to be more consistent across machines.

I think that is a very general problem, and the Axes tool just makes it obvious. If you have drawn axes using the pen previously and tried to align them to the grid, then you would have the same problem when opening that document on another machine with different configuration, wouldn't you?. The question is: what is a centimeter? What is the relation between pixels and centimeters? And OpenBoard answers this depending on the DPI of the display. Different DPI make different grids.
> 
> For the little graphical glitches, I could reproduce it (a tiny bit of the minus sign remains visible, not systematic) but it's not a mandatory thing to fix to make tool axes enabled by default in the next release.

I will definitely look into that, too.
> 
> With these changes made, the tool axes (and the intermediate grid line too) will be enabled by default, and available as-is in the 1.7 :)

kaamui:
> > The last point is imo an important one. I can't see a good way to concile both behaviors, and I see the advantage of the current one, but the fact that the axes and a drawing made above could no longer correspond is an unwanted behavior. If you have a solution to maintain both, I would be happy to hear it. If not, it should be changed to be more consistent across machines.
> 
> I think that is a very general problem, and the Axes tool just makes it obvious. If you have drawn axes using the pen previously and tried to align them to the grid, then you would have the same problem when opening that document on another machine with different configuration, wouldn't you?. The question is: what is a centimeter? What is the relation between pixels and centimeters? And OpenBoard answers this depending on the DPI of the display. Different DPI make different grids.
> 

Yes, totally. However, "stay aligned to the grid" should not be preferred to "stay aligned to the annotations drawn above it".  The same issue occurs with the ruler, but the ruler is not intended to stay on the board with the other objects (at least, less often). I think your axes tool is quite an hybrid object between a ruler and a shape object. This is why it makes this notion of pixel/centimeter ratio you mentioned this obvious. 

I'm not sure I'm explaining my point well so I'll try to translate it in C++ ^^ : I think the behavior wanted by the team is : 
- do `mPixelsPerCentimeter = UBApplication::boardController->activeScene()->backgroundGridSize();` once in the constructor
- don't update mPixelsPerCentimeter in paintGraduations
- persist mPixelsPerCentimeter in the document so it can be loaded when opening the document elsewhere

I didn't test it ^^ it's just to help the discussion

letsfindaway:
Still don't know whether I completely understand the issue. See below.

> The same issue occurs with the ruler, but the ruler is not intended to stay on the board with the other objects (at least, less often). I think your axes tool is quite an hybrid object between a ruler and a shape object. This is why it makes this notion of pixel/centimeter ratio you mentioned this obvious.
> 
> I'm not sure I'm explaining my point well so I'll try to translate it in C++ ^^ : I think the behavior wanted by the team is :
> 
> * do `mPixelsPerCentimeter = UBApplication::boardController->activeScene()->backgroundGridSize();` once in the constructor
> 
> * don't update mPixelsPerCentimeter in paintGraduations

What if the user changes the grid size after dragging the axes to the board? Shouldn't it adapt to the change?
> 
> * persist mPixelsPerCentimeter in the document so it can be loaded when opening the document elsewhere

`backgroundGridSize` of each scene is already persisted as `ub:grid-size` attribute. So after loading the same grid size (pixels per centimeter) should be present. This number is not affected by screen resolution.
> 
> 
> I didn't test it ^^ it's just to help the discussion

So what do I have to do to reproduce the issue? I was still not able to do that.

kaamui:
The feedback I had is this (translated) : 

> So i made further testing, and I just give you an example of what it looks like when I draw a random function above a "Axes" object on Mac, then export it, then import it in another machine (seems related to resolution)
>
Axes créé sur Mac avec f(x) non symétrique

Axes avec f(x) non symétrique sur KUbuntu après import depuis Mac
------------
I must admit I didn't double check if it was reproductible as the images were speaking for themselves. But after I read that you were not able to reproduce it I tried and found that indeed the images were speaking for themselves^^ : I think it happens only if grid is not showed 

My guess is that the guilty line is there : 

https://github.com/OpenBoard-org/OpenBoard/blob/4b040fcf0ed0829c3e2cb260175224a07e23447e/src/adaptors/UBSvgSubsetAdaptor.cpp#L1159-L1165

So I just have to fix this and everything should be OK for the axes to be enabled by default in the next release.

PR #604, opened 2022-05-20
refactor: maintain compatibility with Qt 5.12

letsfindaway:
I now tried to find a way to keep compatibility for a broader range of Qt versions. The only affected class is `UBThumbnailWidget`.

- Make `UBThumbnailWidget` compatible with Qt 5.12 upwards.
- Use a `QList` for member variable `mSelectedThumbnailItems`. This does not affect performance, as it is only used for iterating.
- Convert `QSet` to `QList` using `values()`. This function is not deprecated even in Qt 6.3.

@JBBgameich : do you think this is an appropriate approach?

### DIFF
--- a/src/gui/UBThumbnailWidget.cpp
+++ b/src/gui/UBThumbnailWidget.cpp
@@ -310,8 +310,7 @@ void UBThumbnailWidget::mousePressEvent(QMouseEvent *event)
                 int index2 = mGraphicItems.indexOf(underlyingItem);
                 if (-1 == index2)
                 {
-                    const auto items = selectedItems();
-                    mSelectedThumbnailItems = QSet<QGraphicsItem *>(items.begin(), items.end());
+                    mSelectedThumbnailItems = selectedItems();
                     return;
                 }
                 mSelectionSpan = index2 - index1;
@@ -374,8 +373,8 @@ void UBThumbnailWidget::mouseMoveEvent(QMouseEvent *event)
 
         // for horizontal moving
         const auto selectedItemsX = scene()->items(incrementXSelection, Qt::IntersectsItemBoundingRect);
-        const auto incSelectedItemsX = QSet<QGraphicsItem *>(selectedItemsX.begin(), selectedItemsX.end());
-        for (QGraphicsItem *lassoSelectedItem : incSelectedItemsX)
+
+        for (QGraphicsItem *lassoSelectedItem : selectedItemsX)
         {
             if (lassoSelectedItem)
             {
@@ -401,8 +400,8 @@ void UBThumbnailWidget::mouseMoveEvent(QMouseEvent *event)
         // for vertical moving
 
         const auto selectedItemsY = scene()->items(incrementYSelection, Qt::IntersectsItemBoundingRect);
-        const auto incSelectedItemsY = QSet<QGraphicsItem *>(selectedItemsY.begin(), selectedItemsY.end());
-        for (QGraphicsItem *lassoSelectedItem : incSelectedItemsY)
+
+        for (QGraphicsItem *lassoSelectedItem : selectedItemsY)
         {
             if (lassoSelectedItem)
             {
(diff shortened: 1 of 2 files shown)


JBBgameich:
If this is just about the QSet range constructor, #ifdef QT_VERSION >= QT_VERSION_CHECK(something) ing that line and falling back to the old .toSet() API might be a less invasive approach.

letsfindaway:
> If this is just about the QSet range constructor, #ifdef QT_VERSION >= QT_VERSION_CHECK(something) ing that line and falling back to the old .toSet() API might be a less invasive approach.

That's definitely true. `#if` would be less invasive. However what I don't like about this approach is that then there are always some lines which cannot be tested. Is the code in the `#else` branch really behaving identically to the `#if`? Of course in this very concise case that would not be a problem. But this is the reason why I prefer to avoid such preprocessor directives whenever possible.

All other code is compatible with at least Qt 5.12 and Qt 5.15. And let's see how big the gap to Qt 6.2 will be. Even there, compatible choices are my preference.

letsfindaway:
I was again looking deeper into this piece of code: What does it actually do, what is it used for?

As the name suggests, `mSelectedThumbnailItems` keeps a list/set of selected thumbnail items. In line 313 it is e.g. initialized from `selectedItems()`. If you look into that function, then it returns a **sorted** list of items. Putting that into a `QSet` renders sorting useless.

The original code starting at line 442 is interesting:

```c++
        if (Qt::ControlModifier & event->modifiers())
        {
            for (int i = 0; i < mSelectedThumbnailItems.count()-1; i++)
            {
                mSelectedThumbnailItems.values().at(i)->setSelected(true);
            }
        }
```

It is looping over a `QSet` using the corresponding `QList` by using `values().at(i)` and the loop condition omits the last item of that list. Note however that a `QSet` has arbitrary order, so just an arbitrary item is omitted?! That's at lease strange, and there is no reason to omit that element. It think for **all** elements of `mSelectedThumbnailItems` `setSelected(true)` should be called. And this loop also indicates that a `QList` would probably have been much more appropriate for that variable than a `QSet`. `QSet`'s features for fast lookup of items are never used on that variable.

Finally, this code is in a block guarded with 

```c++
    if (mLassoRectItem)
    {
```

But currently the lasso selection of items is completely disabled. This seems to date back to 2012, where some changes of this logic have been made, but I was not able to completely track that down. Look at the beginning of the `mouseDownEvent`, where the lasso would be created:

```c++
    mMousePressPos = event->pos();

    UBThumbnailPixmap* sceneItem = dynamic_cast<UBThumbnailPixmap*>(itemAt(mMousePressPos));
    if(sceneItem==NULL)
    {
        event->ignore();
        return;
    }

    mMousePressScenePos = mapToScene(mMousePressPos);
    QGraphicsItem* underlyingItem = itemAt(mMousePressPos);
    UBThumbnail *previousSelectedThumbnail = mLastSelectedThumbnail;

    if (!dynamic_cast<UBThumbnail*>(underlyingItem))
    {
        deleteLasso();

        UBRubberBand rubberBand(QRubberBand::Rectangle);
        QStyleOption option;
        option.initFrom(&rubberBand);

        mPrevLassoRect = QRect();
        mLassoRectItem = new QGraphicsRectItem(0);
        scene()->addItem(mLassoRectItem);
```

In the first part, the event is ignored, if the mouse is pressed outside of any item. But is this not exactly the place where a lasso should be created? And yes: later in ` if (!dynamic_cast<UBThumbnail*>(underlyingItem))` exactly that is checked again: is the mouse press outside of an `UBThumbnail` item? But we never get here as the event was already ignored. So also an `mLassorectItem` is never created and the above-mentioned code is never executed.

I just tried to comment out ignoring of the event, and voila! A nice lasso selection feature appears! However I do not know why it was disabled years ago.

This analysis shows me that it is sometimes worth to be "invasive", trying to understand what the code does and why it is as it is. @kaamui : can you shed some light into that rubber band selection feature for thumbnails?

kaamui:
@letsfindaway No idea. I didn't know it was there and don't know why it has been disabled. Maybe because of a sneaky `for` loop ? ^^

PR #608, opened 2022-05-26
Refactor for Qt 6.2

letsfindaway:
This pull request refactors OpenBoard for compilation with Qt 6.2. It aims to create a project which can be compiled with Qt 5.12, Qt 5.15 and Qt 6.2.

The commits of this pull request each address a specific issue. The most complex refactoring were necessary due to the completely reworked media framework. But also the WebEngine APIs have undergone some restructuring. Most of the other things are minor, but often incompatible API changes.

The branch can be compiled with Qt 6.2. However as I'm currently missing `quazip6` for my platform, the files using `quazip` could not be compiled and I'm currently also not able to test it. This will probably change with the upcoming release of openSUSE Leap 15.4 in June 2022 (I'm using Leap as my development platform).

This pull request is still a draft as long as I'm not able to test and verify it. Any comments and experiences on other platforms are welcome!

### DIFF
--- a/src/gui/UBMagnifer.cpp
+++ b/src/gui/UBMagnifer.cpp
@@ -374,7 +374,7 @@ void UBMagnifier::slot_refresh()
 
 void UBMagnifier::grabPoint()
 {
-    QMatrix transM = UBApplication::boardController->controlView()->matrix();
+    QTransform transM = UBApplication::boardController->controlView()->transform();
     QPointF itemPos = gView->mapFromGlobal(updPointGrab);
 
     qreal zWidth = width() / (params.zoom * transM.m11());
@@ -407,7 +407,7 @@ void UBMagnifier::grabPoint()
 
 void UBMagnifier::grabPoint(const QPoint &pGrab)
 {
-    QMatrix transM = UBApplication::boardController->controlView()->matrix();
+    QTransform transM = UBApplication::boardController->controlView()->transform();
     updPointGrab = pGrab;
     QPointF itemPos = gView->mapFromGlobal(pGrab);
 
--- a/src/gui/UBMagnifer.cpp
+++ b/src/gui/UBMagnifer.cpp
@@ -65,10 +65,10 @@ UBMagnifier::UBMagnifier(QWidget *parent, bool isInteractive)
     mResizeItem = new QPixmap(":/images/resize.svg");
     sChangeModePixmap = new QPixmap();
 
-    qDebug() << "sClosePixmap" << sClosePixmap->size() << endl
-             << "sIncreasePixmap" << sIncreasePixmap->size() << endl
-             << "sDecreasePixmap" << sDecreasePixmap->size() << endl
-             << "mResizeItem" << mResizeItem->size() << endl;
+    qDebug() << "sClosePixmap" << sClosePixmap->size() << '\n'
+             << "sIncreasePixmap" << sIncreasePixmap->size() << '\n'
+             << "sDecreasePixmap" << sDecreasePixmap->size() << '\n'
+             << "mResizeItem" << mResizeItem->size() << '\n';
 
 
     setDrawingMode(UBSettings::settings()->magnifierDrawingMode->get().toInt());
@@ -194,11 +194,11 @@ void UBMagnifier::calculateButtonsPositions()
     sDecreasePixmapButtonRect = QRect(sChangeModePixmapButtonRect.x() - sChangeModePixmap->width() - m_iButtonInterval, size().height() - sDecreasePixmap->height(), sDecreasePixmap->width(), sDecreasePixmap->height());
     sIncreasePixmapButtonRect = QRect(sDecreasePixmapButtonRect.x() - sChangeModePixmap->width() - m_iButtonInterval, size().height() - sIncreasePixmap->height(), sIncreasePixmap->width(), sIncreasePixmap->height());
 
-    qDebug() << "mResizeItemButtonRect" << mResizeItemButtonRect << endl
-             << "sClosePixmapButtonRect" << sClosePixmapButtonRect << endl
-             << "sChangeModePixmapButtonRect" << sChangeModePixmapButtonRect << endl
-             << "sDecreasePixmapButtonRect" << sDecreasePixmapButtonRect << endl
-             << "sIncreasePixmapButtonRect" << sIncreasePixmapButtonRect << endl;
+    qDebug() << "mResizeItemButtonRect" << mResizeItemButtonRect << '\n'
+             << "sClosePixmapButtonRect" << sClosePixmapButtonRect << '\n'
+             << "sChangeModePixmapButtonRect" << sChangeModePixmapButtonRect << '\n'
+             << "sDecreasePixmapButtonRect" << sDecreasePixmapButtonRect << '\n'
+             << "sIncreasePixmapButtonRect" << sIncreasePixmapButtonRect << '\n';
 }
 
 void UBMagnifier::paintEvent(QPaintEvent * event)
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -33,13 +33,12 @@ VERSION_RC = $$replace(VERSION_RC, "rc", "192" ) # 0xC0
 VERSION_RC = $$replace(VERSION_RC, "r", "240") # 0xF0
 
 QT += svg
+greaterThan(QT_MAJOR_VERSION, 5):QT += svgwidgets
 QT += network
 QT += xml
-QT += xmlpatterns
 QT += uitools
 QT += multimedia
 QT += multimediawidgets
-QT += webengine
 QT += webenginewidgets
 QT += printsupport
 QT += core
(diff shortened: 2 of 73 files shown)


JBBgameich:
Nice work!

With these changes I could successfully run it on Qt 6.3.0 including the quazip parts (I compiled QuaZip myself).
```diff
diff --git a/src/adaptors/UBImportCFF.cpp b/src/adaptors/UBImportCFF.cpp
index f8e1d679..79e6bad1 100644
--- a/src/adaptors/UBImportCFF.cpp
+++ b/src/adaptors/UBImportCFF.cpp
@@ -272,7 +272,7 @@ UBDocumentProxy* UBImportCFF::importFile(const QFile& pFile, const QString& pGro
         dir.mkdir(destDocument->persistencePath());
         if (pGroup.length() > 0)
             destDocument->setMetaData(UBSettings::documentGroupName, pGroup);
-        if (fi.baseName() > 0)
+        if (!fi.baseName().isEmpty())
             destDocument->setMetaData(UBSettings::documentName, fi.baseName());
 
         destDocument->setMetaData(UBSettings::documentVersion, UBSettings::currentFileVersion);
diff --git a/src/adaptors/UBImportDocumentSetAdaptor.cpp b/src/adaptors/UBImportDocumentSetAdaptor.cpp
index 5ee3b3b2..4a4ea95a 100644
--- a/src/adaptors/UBImportDocumentSetAdaptor.cpp
+++ b/src/adaptors/UBImportDocumentSetAdaptor.cpp
@@ -105,7 +105,7 @@ QFileInfoList UBImportDocumentSetAdaptor::importData(const QString &zipFile, con
 
         QString newFilePath = destination + "/" + newFileName;
         if (UBFileSystemUtils::copy(readDir.absoluteFilePath(), newFilePath, true)) {
-            result.append(newFilePath);
+            result.append(QFileInfo(newFilePath));
         }
     }
```

On startup, there's a number of connect warnings which likely mean some connects are no longer working and need the target function signature updated. I think ideally the connects should be switched to the newer compile-time checked syntax, which would avoid this kind of errors. I can try to provide a pull request for that.
```
QObject::connect: No such signal QButtonGroup::buttonClicked(int)
QObject::connect: No such signal QButtonGroup::buttonClicked(int)
QObject::connect: No such signal QButtonGroup::buttonClicked(int)
QObject::connect: No such signal QButtonGroup::buttonClicked(int)
QObject::connect: No such signal QButtonGroup::buttonClicked(int)
QObject::connect: No such signal QComboBox::currentIndexChanged(const QString &)
QObject::connect:  (sender name:   'keyboardPaletteKeyButtonSize')
```

JBBgameich:
I opened a pull request on your repository which fixes the mentioned issues. If you merge it, the commits should appear in this pull request

src/gui/UBMagnifer.cpp
```diff
@@ -194,11 +194,11 @@ void UBMagnifier::calculateButtonsPositions()
     sDecreasePixmapButtonRect = QRect(sChangeModePixmapButtonRect.x() - sChangeModePixmap->width() - m_iButtonInterval, size().height() - sDecreasePixmap->height(), sDecreasePixmap->width(), sDecreasePixmap->height());
     sIncreasePixmapButtonRect = QRect(sDecreasePixmapButtonRect.x() - sChangeModePixmap->width() - m_iButtonInterval, size().height() - sIncreasePixmap->height(), sIncreasePixmap->width(), sIncreasePixmap->height());
 
-    qDebug() << "mResizeItemButtonRect" << mResizeItemButtonRect << endl
-             << "sClosePixmapButtonRect" << sClosePixmapButtonRect << endl
-             << "sChangeModePixmapButtonRect" << sChangeModePixmapButtonRect << endl
-             << "sDecreasePixmapButtonRect" << sDecreasePixmapButtonRect << endl
-             << "sIncreasePixmapButtonRect" << sIncreasePixmapButtonRect << endl;
+    qDebug() << "mResizeItemButtonRect" << mResizeItemButtonRect << '\n'
```
JBBgameich, line 197: Replacing endl with '\n' is not necessary, it has simply been moved into the Qt namespace to not conflict with std::endl.
So Qt::endl would work.

letsfindaway: Seems that I do not use the right search terms, but I cannot find the definition of `Qt::endl` either in the Qt documentation nor by grepping the include directories. Can you help me?

And I remember that again for my old Qt 5.12 this definition of `Qt::endl` was not available, so I chose the "compatible" way.

JBBgameich: Here is the documentation for it: https://doc.qt.io/qt-5/qt.html#endl
Seems like it was introduced in Qt 5.14.

letsfindaway: Yeah, I had to read some pages explaining how stream modifiers actually work and how just specifying their name invokes them on the stream, but I hope I now have understood that. And yes, your porposal looks fine. Will look on it tomorrow. Note that I'm on the openSUSE conference in Nuremberg from Wednesday to Saturday, so I might not be responsive during this time.

letsfindaway: My approach would now be to just define the missing `Qt::endl` for Qt < 5.14:
```c++
#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
namespace Qt 
{
inline QTextStream &endl(QTextStream &s) {
    return ::endl(s);
}
#endif
```
When dropping Qt < 5.14 compatibility just this definition will be dropped, the code needs no adaptation.

kaamui: According to [flush's documentation](https://doc.qt.io/qt-5/qtextstream.html#flush), if QTextStream operates on a string, flush does nothing, so I don't think there is a slight performance win on these cases. But preprocessor  directives seem avoidable with endl doing `return stream << '\n'_L1 << Qt::flush;` ([source](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/serialization/qtextstream.cpp?h=dev)). 

So I would personally prefer @letsfindaway's solution. I just wonder what's the reason to use a QChar1Latin. Is it related to `Note: On Windows, all '\\n' characters are written as '\\r\\n' if
    QTextStream's device or string is opened using the QIODevice::Text flag.` described in endl's documentation ?

letsfindaway: We're operating on I/O streams (with `qDebug()`) or files (in `UBGraphicsWidgetItem.cpp:1146ff`), so flushing affects performance.

And for the `_L1` appendix: this is rather new, I thing with upcoming Qt 6.4, so we don't have to care about it currently. I think it just makes clear that this character is 8 bits and shall remain 8 bits.

kaamui:
oops sry I read too fast the line you had pointed and just saw `"<html>..." << endl`

src/gui/UBMagnifer.cpp
kaamui: oops sry I read too fast the line you had pointed and just saw `"<html>..." << endl`

kaamui: > We're operating on I/O streams (with `qDebug()`) ... so flushing affects performance.

I thought qDebug() calls were totally ignored in release mode, but now that I think of it I don't know how exactly. Is there a way they would be ignored or go to a "do nothing" directive but the flush still performed on an empty stream ? Would this affect performance in a release context ? If not, it should not bother us that much...

> And for the `_L1` appendix: this is rather new, I thing with upcoming Qt 6.4, so we don't have to care about it currently. I think it just makes clear that this character is 8 bits and shall remain 8 bits.

The `_L1` syntax could be rather new, but there was a QLatin1Char instanciation instead in the previous Qt versions (I suppose the "character litteral" syntax  just saves some typing ?), so I still wonder why it is done. Could it be important to maintain the use of `endl` on files to handle OS-specific end of lines ?

letsfindaway: I still assume that it does not matter. What happens:

- using `<< '\n'` the `QTextStream &operator<<(char ch)` will be used. Internally this is implemented as:
```c++
QTextStream &QTextStream::operator<<(char c)
{
    Q_D(QTextStream);
    CHECK_VALID_STREAM(*this);
    d->putChar(QChar::fromLatin1(c));
    return *this;
}
```
So it interprets the argument as Latin1, converts it to a `QChar` and then puts it to the stream.

- using `QLatin1Char('\n')` there is no explicit stream operator for that type, so the implicit conversion operator of `QChar` will be used to convert it and then use `QTextStream &operator<<(QChar ch)`. You can imagine how this implementation looks like:
```c++
QTextStream &QTextStream::operator<<(QChar c)
{
    Q_D(QTextStream);
    CHECK_VALID_STREAM(*this);
    d->putChar(c);
    return *this;
}
```

In both cases the same conversion steps are performed. The handling of the platform specific line endings is handled in the underlying `QIODevice`, not in `QTextStream`.

kaamui:
With Qt 6.4.0 => https://bugreports.qt.io/browse/QTBUG-105984 (fixed in Qt 6.4.1)

letsfindaway:
Removed most of the changes to the project file except those explicitly dealing with Qt 6.

PR #620, opened 2022-06-05
The show and hide events in UBDesktopPalette have no purpose.

habruening:
I think, these two events have no purpose. The events are activated when Open Board is minimised by the window manager. But there are a few things, with this code:
* Open Board in the desktop mode cannot be minimised.
* It does not matter if in minimsed mode the arrorws are there or not.
* I see no reason at all why in certain situations the arrows should not be invisble. I could imagine that the author had something like a kiosk mode in mind or some things that are specific to certain use cases. But at the moment I don't see such a reason. If there is a reason, we should have a comment in the code.

### DIFF
--- a/src/desktop/UBDesktopPalette.cpp
+++ b/src/desktop/UBDesktopPalette.cpp
@@ -208,42 +208,6 @@ void UBDesktopPalette::maximizeMe()
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


habruening:
Thanks for the clarification. That was not so obvious.

PR #622, opened 2022-06-05
The window capture function is not used any more.

habruening:
I don't see that the windows capture function is ever used. It looks like it has been replaced with the custom capture function. I removed the windows capture function. As far as I can see everything still works.
If I am right, have here a pull request.


letsfindaway:
Yes, you're right. However this is already done in the current `1.7-dev` branch with commit https://github.com/OpenBoard-org/OpenBoard/commit/1cda944ea9d802e20d4f041bb2a895db540f2ddb. See also https://github.com/letsfindaway/OpenBoard/issues/74#issuecomment-1113985863 (last paragraph).

habruening:
Thanks for the information. So I will switch to 1.7-dev.

PR #627, opened 2022-06-10
Fix syntax error message when executing qmake

maehne:
On Debian 11 when executing `qmake OpenBoard.org`, one gets the following
error message:

```
qmake OpenBoard.pro
sh: 1: Syntax error: "&" unexpected
sh: 1: Syntax error: "&" unexpected
sh: 1: Syntax error: "&" unexpected
```

The cause is the use of bash-specific `|&` pipe operator in `podcast.pri` line 70
to determine the `FFMPEG_VERSION`.

### DIFF
--- a/src/podcast/podcast.pri
+++ b/src/podcast/podcast.pri
@@ -67,7 +67,7 @@ linux-g++* {
             -lxcb-xfixes \
             -lxcb-render -lxcb-shape -lxcb -lX11 -lasound -lSDL -lx264 -lpthread -lvpx -lvorbisenc -lvorbis -ltheoraenc -ltheoradec -logg -lopus -lmp3lame -lfreetype -lfdk-aac -lass -llzma -lbz2 -lz -ldl -lswresample -lswscale -lavutil -lm
 
-    FFMPEG_VERSION = $$system(ffmpeg --version|& grep -oP "version.*?\K[0-9]\.[0-9]")
+    FFMPEG_VERSION = $$system(ffmpeg --version 2>&1 | grep -oP "version.*?\K[0-9]\.[0-9]")
     equals(FFMPEG_VERSION, 2.8) {
         LIBS -= -lswresample
         LIBS += -lavresample


letsfindaway (commented):
Thanks for this PR. I can approve that is solves the stated issue in environments where `bash` is not the standard shell.

However I want to ask @kaamui whether support for such old ffmpeg versions is still required. We currently are on ffmpeg 4 and will probably be moving to ffmpeg 5 in the near future. In my proposal for compatibility with Qt 6 I was proposing in commit https://github.com/letsfindaway/OpenBoard/commit/d26bb19e0a6cbb71b27577759c0cc5c9e693ed93 to completely replace this library configuration by `pkgconfig` without version check. This makes sure that the correct libraries and include paths are chosen independent of version or target platform, but would break compatibility with ffmpeg 2.8.

maehne:
@letsfindaway: Thanks for the positive review! I am all for dropping support for very old versions of dependencies if it simplifies the code and renders building more robust. Nevertheless, I would appreciate if your fix or my fix to the issue would soon land on 1.7-dev to be able to reduce the number of local patches to maintain.

mlo-poc:
I had to chande another detail:

the `--version` option was not recognized, so i had to change it to `-version` (just one dash)
```
diff --git a/src/podcast/podcast.pri b/src/podcast/podcast.pri
index 873658c9..61099c4c 100644
--- a/src/podcast/podcast.pri
+++ b/src/podcast/podcast.pri
@@ -67,7 +67,7 @@ linux-g++* {
             -lxcb-xfixes \
             -lxcb-render -lxcb-shape -lxcb -lX11 -lasound -lSDL -lx264 -lpthread -lvpx -lvorbisenc -lvorbis -ltheoraenc -ltheoradec -logg -lopus -lmp3lame -lfreetype -lfdk-aac -lass -llzma -lbz2 -lz -ldl -lswresample -lswscale -lavutil -lm
 
-    FFMPEG_VERSION = $$system(ffmpeg --version|& grep -oP "version.*?\K[0-9]\.[0-9]")
+    FFMPEG_VERSION = $$system(ffmpeg -version | grep -oP "version.*?\K[0-9]\.[0-9]")
     equals(FFMPEG_VERSION, 2.8) {
         LIBS -= -lswresample
         LIBS += -lavresample

```

PR #633, opened 2022-06-15
Improve displaymanager

letsfindaway:
This pull request continues work on the `UBDisplayManager` and related classes. I want to thank all discussion partners and testers for their input and hope that most of them found its way to this PR.

The most noticeable improvement is the screen configuration by index instead of name. Names are too platform specific and volatile, so a simple index is the better choice. But much more has changed. Here a list in the order as the files appear below:

### UBBoardController

- The background grid size is calculated when the display manager has defined the roles of the screens.
- The boxing in window mode is calculated as if the control screen would be full screen.
- Boxing is recalculated when the window is resized.

### UBBoardView

Wheel events on "Previous" screens are no longer handled.

### UBApplicationController

- The initialization and adjustment of previous screens is now dynamically adapting to the screen configuration

### UBDisplayManager

- Screen indexes, roles and positions are now processed in three separate phases.
- Each phase is linked to the proper events and also emits an event when finished.
- This avoids duplicate processing and makes the timed positioning (hopefully) unnecessary, except for attaching or removing a monitor.
- The configurable screen list now contains a list of indexes.
- A lot of formatting and style.
- Improved pointer checks.

### UBPersistenceWorker

- Simple typo.

### PreferencesController

- As explained above, use a list of indexes for configuration.
- Store the screen list per monitor configuration.
- Remove the `QCompleter`, it makes no sense for indexes.
- Don't allow the screen labels to take focus.
- No invalid input, but make incomplete input user visible with background color.
- Remove the fade-out timer, it is no longer necessary.
- Emit signal when screen list changed.

### UBPlatformUtils (all platforms)

- Apply window mode to control widget only.

### UBDockPalette

- Reposition docks when screen was resized.

### DIFF
--- a/src/core/UBPreferencesController.cpp
+++ b/src/core/UBPreferencesController.cpp
@@ -38,6 +38,8 @@
 #include "core/UBApplicationController.h"
 #include "core/UBDisplayManager.h"
 
+#include "frameworks/UBStringUtils.h"
+
 #include "board/UBBoardController.h"
 #include "document/UBDocumentController.h"
 #include "domain/UBGraphicsScene.h"
@@ -57,21 +59,6 @@ typedef QString::SplitBehavior SplitBehavior;
 qreal UBPreferencesController::sSliderRatio = 10.0;
 qreal UBPreferencesController::sMinPenWidth = 0.5;
 qreal UBPreferencesController::sMaxPenWidth = 50.0;
-
-// convenience function to split a comma separated list of tokens to a QStringList
-QStringList trimmedTokens(const QString& input)
-{
-    QStringList tokens = input.split(',');
-
-    for (QString& entry : tokens)
-    {
-        entry = entry.trimmed();
-    }
-
-    tokens.removeAll("");
-
-    return tokens;
-}
 
 
 UBPreferencesDialog::UBPreferencesDialog(UBPreferencesController* prefController, QWidget* parent,Qt::WindowFlags f)
@@ -821,7 +808,7 @@ void UBScreenListLineEdit::focusInEvent(QFocusEvent *focusEvent)
 
     if (mScreenLabels.empty())
     {
-        QStringList screenList = trimmedTokens(text());
+        QStringList screenList = UBStringUtils::trimmed(text().split(','));
 
         QList<QScreen*> screens = UBApplication::displayManager->availableScreens();
         QStringList availableScreenIndexes;
@@ -892,7 +879,7 @@ void UBScreenListLineEdit::addScreen()
 
 void UBScreenListLineEdit::onTextChanged(const QString &input)
 {
-    QStringList screenList = trimmedTokens(input);
+    QStringList screenList = UBStringUtils::trimmed(input.split(','));
 
     for (QPushButton* button : mScreenLabels)
     {
@@ -934,7 +921,7 @@ UBStringListValidator::UBStringListValidator(QObject *parent)
 void UBStringListValidator::fixup(QString &input) const
 {
     // remove invalid tokens from list, trim tokens
-    QStringList inputList = trimmedTokens(input);
+    QStringList inputList = UBStringUtils::trimmed(input.split(','));
     QStringList outputList;
 
     for (const QString& token : inputList)
@@ -951,7 +938,7 @@ void UBStringListValidator::fixup(QString &input) const
 QValidator::State UBStringListValidator::validate(QString &input, int &) const
 {
     bool ok = true;
-    QStringList inputList = trimmedTokens(input);
+    QStringList inputList = UBStringUtils::trimmed(input.split(','));
     // number of commas must match number of list items - 1
     int commas = input.count(',');
 
--- a/src/core/UBPreferencesController.cpp
+++ b/src/core/UBPreferencesController.cpp
@@ -58,6 +58,21 @@ qreal UBPreferencesController::sSliderRatio = 10.0;
 qreal UBPreferencesController::sMinPenWidth = 0.5;
 qreal UBPreferencesController::sMaxPenWidth = 50.0;
 
+// convenience function to convert between screen list as QString and QStringList
+QStringList splitScreenList(const QString& input)
+{
+    QStringList screenNames = input.split(',');
+
+    for (QString& entry : screenNames)
+    {
+        entry = entry.trimmed();
+    }
+
+    screenNames.removeAll("");
+
+    return screenNames;
+}
+
 
 UBPreferencesDialog::UBPreferencesDialog(UBPreferencesController* prefController, QWidget* parent,Qt::WindowFlags f)
     :QDialog(parent,f)
@@ -131,7 +146,7 @@ void UBPreferencesController::wire()
     connect(mPreferencesUI->screenList, &QLineEdit::textChanged, [this,settings](const QString &text){
         if (mPreferencesUI->screenList->hasAcceptableInput())
         {
-            settings->appScreenList->set(text.split(','));
+            settings->appScreenList->set(splitScreenList(text));
         }
     });
 
@@ -784,7 +799,7 @@ void UBScreenListLineEdit::focusInEvent(QFocusEvent *focusEvent)
 
     if (mScreenLabels.empty())
     {
-        QStringList screenNames = text().split(',', SplitBehavior::SkipEmptyParts);
+        QStringList screenNames = splitScreenList(text());
 
         QList<QScreen*> screens = UBApplication::displayManager->availableScreens();
         QStringList availableScreenNames;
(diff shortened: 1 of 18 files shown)


habruening:
The name UBStringListValidator does not fit well. It is not a general String list validator.

UBScreenListLineEdit::mValidator does not have a parent. I think, Qt will not consider UBScreenListLineEdit as the owner of it and therefore does not delete it for you. Setting setValidator does not take ownership.

I would not pass availableScreenIndexes to the constructor. And I would not always create a fresh validator. The validator can call UBApplication::displayManager->availableScreens() on its own because it is static.

letsfindaway:
Hi @habruening ,

thanks for the review!

> The name UBStringListValidator does not fit well. It is not a general String list validator.

Hmm, what is it? It validates the following:

- The input must be a comma-separated list of zero or more items.
- Leading and trailing whitespace around each item are ignored.
- Each item in the list must be unique and must match one of the items of the validator's list.

So for me it is pretty much a general string list validator, with some additional rules for whitespace and uniqueness. Anyway: do you have a better proposal?

> UBScreenListLineEdit::mValidator does not have a parent. I think, Qt will not consider UBScreenListLineEdit as the owner of it and therefore does not delete it for you. Setting setValidator does not take ownership.

Yep, thanks. I will set a parent.

> I would not pass availableScreenIndexes to the constructor. And I would not always create a fresh validator. The validator can call UBApplication::displayManager->availableScreens() on its own because it is static.

That's a design question. If I would let the validator ask for the list, then your first point is definitely true: Then it would not be a general string list validator, but a class which is only usable in a very specific context. I always try to avoid this. There is no need for the validator to have a reference to the display manager. 

Still two options remain:

- I could set the string list in the constructor with no means to later change it.
- I could set the string list via some `setValidationStringList` function.

The first option creates an immutable, stateless object. It is guaranteed, that during the lifetime of this object every call to `validate` or `fixup` will behave the same. It has some more advantages, e.g. it is thread-safe, but none of them seems to be important in this specific situation. The second option creates a more flexible object, where the validation model can be changed during lifetime. The advantage is that it does not have to be recreated each time.

For me the pro's and con's do not heavily weight in this case. And it would probably more Qt style to make it mutable. Qt does not have many immutable objects (`QRegularExpression` is one due to performance reasons, but this does not count here). So I would probably go and add a setter for the list as proposed above.

habruening:
- Now I see. You are right, UBStringListValidator is general. I was confused because of the call of splitScreenList. This name makes the code look very specific. But I was wrong. You could indeed rename this function. Maybe you don't need it, at all. There is no need of going through the list twice. The code of fixup can already trim strings and skip the empty ones.
- And then of course it makes sense not calling something in displayManager.

src/core/UBPreferencesController.cpp
```diff
@@ -52,19 +52,19 @@ qreal UBPreferencesController::sSliderRatio = 10.0;
 qreal UBPreferencesController::sMinPenWidth = 0.5;
 qreal UBPreferencesController::sMaxPenWidth = 50.0;
 
-// convenience function to convert between screen list as QString and QStringList
-QStringList splitScreenList(const QString& input)
+// convenience function to split a comma separated list of tokens to a QStringList
+QStringList trimmedTokens(const QString& input)
```
kaamui, line 56: I (personnally) prefer when function and variable names are explicit on what it represents in functional terms rather than in a technical aspect. Maybe something like `getScreenNamesFromInput` or `toScreenNames` or `convertToScreenNames` with the idea to have something like `auto screenNames = extractScreenNames(input); for (auto screeName : screenNames) ...` when using it. 

for example, I would expect from a function like `splitScreenList` to have a signature like `QList<QScreen*> splitScreenList(const QString& input)`. For `trimmedTokens`, I would expect something like `void trimmedTokens(const QStringList& tokens)`

edit : `screenIndexes` is probably even more accurate

src/core/UBPreferencesController.h
```diff
@@ -152,20 +155,22 @@ private slots:
 
 private:
     QList<QPushButton*> mScreenLabels;
-    QValidator* mValidator;
+    UBStringListValidator* mValidator;
 };
 
 class UBStringListValidator : public QValidator
```
kaamui, line 161: For clarity/accuracy reasons, this class should be more explicit and be renamed `UBScreenNamesValidator`

src/core/UBPreferencesController.cpp
```diff
@@ -918,17 +917,16 @@ void UBScreenListLineEdit::onTextChanged(const QString &input)
     }
 }
 
-UBStringListValidator::UBStringListValidator(QStringList list, QObject *parent)
+UBStringListValidator::UBStringListValidator(QObject *parent)
     : QValidator(parent)
-    , mList(list)
```
kaamui, line 891: same idea, `mList` should be renamed `screenNames`

src/core/UBPreferencesController.cpp
```diff
@@ -884,11 +884,11 @@ void UBScreenListLineEdit::addScreen()
 
 void UBScreenListLineEdit::onTextChanged(const QString &input)
 {
-    QStringList screenNames = splitScreenList(input);
```
kaamui: `screenNames` was better ! ^^

src/core/UBPreferencesController.cpp
```diff
@@ -829,27 +832,24 @@ void UBScreenListLineEdit::focusInEvent(QFocusEvent *focusEvent)
...
         {
-            setValidator(nullptr);
-            delete mValidator;
+            mValidator = new UBStringListValidator(this);
+            setValidator(mValidator);
         }
 
-        mValidator = new UBStringListValidator(availableScreenIndexes);
-        setValidator(mValidator);
+        mValidator->setValidationStringList(availableScreenIndexes);
```
kaamui, line 852: `setValidScreenNames` ? Or `setAcceptableScreenNames` ? Or maybe even more accurate, `setAcceptableScreenIndexes`, with the idea to replace every mention of `screenNames` with `screenIndexes`

src/core/UBPreferencesController.cpp
letsfindaway: > I (personnally) prefer when function and variable names are explicit on what it represents in functional terms rather than in a technical aspect.

Here I might have the opposite (personal) preference ;) My argument would be that a function operating on a string and returning a somehow splitted version of it has no idea that the content of that string has something to do with screens and should therefore not have "screen" in its name. It is the **invoker** who uses it for the purpose of splitting a screen index list. The function could also be used to split a list of domain names, options, or whatever in another context. Just think you move this function out of this source file to some common string utility place. Then it would be strange if it still contained "screen" in its name. And indeed, would perhaps `UBTextTools` also be a good location for this function?

In the end the function mentioned above is some specialized version of `QString::split`, so I could imagine things like
- `splitAndTrim`
- `splitStringAndTrim`
- `splitToTrimmedTokens`

src/core/UBPreferencesController.h
letsfindaway: For this and everything below see also the discussion with @habruening above. I made some effort to make the `UBStringListValidator` completely unaware of what it validates. It just checks whether the input string can be split into trimmed tokens and that then each of the tokens is contained in the validation string list and occurs at most once. 

E.g. I had not to change anything in the validator when moving from string names to string indexes. It is completely unaware of that. And as with the string splitter above, it could also be used on another completely unrelated input field containing e.g. a comma separated list of options for something completely different. So again I would argue that it should not have "screen" as part of its name or any of its members.

src/core/UBPreferencesController.cpp
kaamui: Oh, I totally agree IF the function is generic but then, indeed its location should be in a class responsible of manipulating strings like `UBStringUtils` in the frameworks folder.

Now I understand that you were trying to make your new classes reusable, like with UBStringListValidator. It should also be located in `UBStringUtils`, imo.

If the intent is to make these functions more generic, I think it could also respect SRP. My experience is that when I need to put an `and` in a function's name, either I didn't find the correct word explaining its responsibility, or I should have make two functions. 

Maybe having something like `static QStringList UBStringUtils::toStringList(const QString& source, const QChar separator) ` and `static QStringList UBStringUtils::trim(const QStringList& list)` would be more accurate and reusable (and it would respect SRP). What do you think ?

letsfindaway: Very good idea! Agreed!

habruening: In my opinions everybody wants to be generic and wants to make things reusable. But then the situation never comes, where something is actually reused. UBPreferenceController is no good location to define such general functions, and it is no good location to search for such general functions. If the UBStringListValidator was really something that is worth having it in general, it would be part of Qt or of some utility library.
Therefore I always prefer to use names that represent functional aspects. This enables an abstract way of programming. And this results in code that is nice to read and to understand. Code is typically written once, but read multiple times. Technical aspects should not be mixed with functional aspects and should not distract from them. Using strings is just an unimportant technical detail and therefore should not influence how you name your functions. Someone may even question if using strings is a good idea. In my opinion it does not matter if you use strings or not. If you change the internal data representation to something else, the code should not have to be changed, because the technical details should be encapsulated away from your code.

That is just my opinion. But I always respect the decion of the author.

letsfindaway: See commit https://github.com/OpenBoard-org/OpenBoard/pull/633/commits/8d5d76a6a56a7e50cbef1acbe0cb6655eacddfd5 as the result of the discussion above. 

Still the `UBStringListValidator` is part of the preferences controller. You are right, one would not expect a reusable class here. For me however there is another reason for making classes in a reusable way: it typically also improves "separation of concerns", and this is an argument even if in the end nobody actually reuses the code.

PR #634, opened 2022-06-15
Desktop Palette disappears before screen capture

habruening:
There is a function in Open Board that the desktop palette disappears before screen capture. This does not work on my System (Debian Linux). I made a fix, that works. Please don't consider it as a pull request yet. Before continuing with it I need your feedback. There are a few things that I want to have clarified before completing the fix:

1. I think, with my fix, the variable mIsFullyTransparent and the logic of it is no longer required. I would remove it. But I would like to know If you see anything against it?
2. I think, I also don't have to update the background any more.
3. Would my lambda function an acceptable programming style? Not everybody likes them. Of course the formatting would be different. I did it so that you can compare it with the original code.
4. I think, the 1 second timer will be no 100% reliable solution. We would need a synchronization with the hideEvent, so that we can do the capture right after the widget is hidden. I can do this. But on the other hand, waiting 1 second is not such a bad solution.

### DIFF
--- a/src/desktop/UBDesktopAnnotationController.cpp
+++ b/src/desktop/UBDesktopAnnotationController.cpp
@@ -448,8 +448,8 @@ void UBDesktopAnnotationController::customCapture()
 
     if (customCaptureWindow.execute(getScreenPixmap()) == QDialog::Accepted)
     {
-       QPixmap selectedPixmap = customCaptureWindow.getSelectedPixmap();
-       emit imageCaptured(selectedPixmap, false);
+        QPixmap selectedPixmap = customCaptureWindow.getSelectedPixmap();
+        emit imageCaptured(selectedPixmap, false);
     }
 
     mDesktopPalette->show();
--- a/src/desktop/UBDesktopAnnotationController.cpp
+++ b/src/desktop/UBDesktopAnnotationController.cpp
@@ -435,24 +435,28 @@ void UBDesktopAnnotationController::customCapture()
     mIsFullyTransparent = true;
     updateBackground();
 
-    mDesktopPalette->disappearForCapture();
-    UBCustomCaptureWindow customCaptureWindow(mDesktopPalette);
-    // need to show the window before execute it to avoid some glitch on windows.
+    mDesktopPalette->hide();
 
-#ifndef Q_OS_WIN // Working only without this call on win32 desktop mode
-    UBPlatformUtils::showFullScreen(&customCaptureWindow);
-#endif
+    QTimer::singleShot(1000, [this]() {
 
-    if (customCaptureWindow.execute(getScreenPixmap()) == QDialog::Accepted)
-    {
-        QPixmap selectedPixmap = customCaptureWindow.getSelectedPixmap();
-        emit imageCaptured(selectedPixmap, false);
-    }
+        UBCustomCaptureWindow customCaptureWindow(mDesktopPalette);
+        // need to show the window before execute it to avoid some glitch on windows.
 
-    mDesktopPalette->appear();
+    #ifndef Q_OS_WIN // Working only without this call on win32 desktop mode
+        UBPlatformUtils::showFullScreen(&customCaptureWindow);
+    #endif
 
-    mIsFullyTransparent = false;
-    updateBackground();
+        if (customCaptureWindow.execute(getScreenPixmap()) == QDialog::Accepted)
+        {
+            QPixmap selectedPixmap = customCaptureWindow.getSelectedPixmap();
+            emit imageCaptured(selectedPixmap, false);
+        }
+
+        mDesktopPalette->show();
+
+        mIsFullyTransparent = false;
+        updateBackground();
+    });
 }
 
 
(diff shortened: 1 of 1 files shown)


letsfindaway:
Hi,

indeed, hiding the desktop palette for screen shots also does not work for me.

I just have some questions to understand your choices:

- Did you find out why the original code in `disappearForCapture` and `appear` does no longer work? But transparency seems to rely on the underlying window manager, so replacing that is probably anyway a good idea.
- What would be needed to make the fix in `disappearForCapture` and `appear`? I think if these are the functions currently failing, then it would be nice if they could be fixed instead of adding a workaround at the caller side.
- For me the mouse cursor is still an arrow when entering the area of the palette, even if it is hidden. Did you also observe this and have you any idea why this is the case? Would be nice if this could also be fixed!
- When using `hide` and `show` in `disappearForCapture` and `appear` you could also make sure that the associated event handlers are skipped and do not the unnecessary work of modifying the action icons.
- One idea would be that `disappearForCapture` returns a `QFuture` which is resolved after some time. A pattern which worked for me was:

```c++
QFuture<void> UBDesktopPalette::disappearForCapture()
{
    hide();

    return QtConcurrent::run([](){
        // delay some time
        QThread::msleep(100);
    });
}

// and at the caller

mDesktopPalette->disappearForCapture().waitForFinished();
```

This moves the timer to the callee, but still executes the delay in the caller. But just as an idea...

habruening:
1. Thanks for the feedback. I think, there are two problems with `disappearForCapture` and `appear`. They use transparency. I did not find out why this does not work on my system. And they don't give the widget enough time to disappear.
Of course, we will not workaround things. The question is how can we fix `disappearForCapture` and `appear`, so that the caller can just use them.

2. A future is no invention of Qt. In this case it is not the right solution. A future is for computations that should be done in parallel, while we can do other things. A future is like a variable that gets its value not immediately, but in the future. A very typical example is to start a complex computation in a constructor. Typically the member variables are defined in the constructor but they are used in the methods later. If it takes longer time to calculate one of the member variables, this can be done in parallel. The constructor can return immediately and let the computation run. The result of the calculation can be used then in the method calls, hoping that the computation is then complete.

Here we have no use case for a future. We have no computation, and we don't want to do something while waiting. And we need no result.

Unfortunately when passing the solution to `disappearForCapture` or use a future, the code will look nearly identical to what I already have. There will be a lambda and the rest of `UBDesktopAnnotationController::customCapture` will be in that lambda. 

The only alternative I found is another slot like `UBDesktopAnnotationController::doCustomCapture`. But that blows up the code and pollutes even the .h file. This is against the practice "do things as local as possible".

3. Removing the code for modifying action icons is not my concern at the moment. The current mechanism is weird. It is absolutely not expected that calling hide and show has this exotic side effect. Side side effects should be avoided. The author of this must have had in mind that this this mechanism engages inadvertently. And therefore I can live with it now. I think, this should be improved. But that is another topic, that I want to do independently.

4. Another question is if we are happy with a time delay. The window manager may have some fade out effects, so that hiding takes 2 seconds. But a time delay of 2 seconds is not acceptable. Then the question is what can be the timeout? I think a sychronization with the completion of the hide Event is the better approach. But as we already see, the code with the lambda and the timeout is already so complicated. I don't know if it is worth making it even more complicated.

But anyway. I will continue with it and come with another draft in a few weeks. Thanks for the input.

letsfindaway:
Sorry for my proposal with the future, it was completely nonsense here, and I already removed it. Must have been too much sun or whatever ;) I completely agree to your analysis above.

> 4\. I think a synchronization with the completion of the hide Event is the better approach.

Do you have an idea how to accomplish this? Is there any feedback that the fade-out has completed?

I did not think of graphical effects taking so long. As I don't have enabled any of them, 100ms works for me in any case. But I see this might vary.

letsfindaway:
Above I made a remark that the mouse cursor changes it it is over the area of the - now hidden - desktop palette. The reason is that the widget mask, which explicitly excludes palette areas from the widget, is still active and has a "hole" at this position.

This also brought me to the fact, that even the right palette (if you made it visible before screenshot) does not disappear and even is still functional at the moment where you're supposed to make your selection of the screen shot rectangle.

So probably more things about screenshot and palettes would benefit from some improvement.

habruening:
Thinks for the comments.
Now I am trying to find a good solution for hiding the widget before capture. I read in Internet that after the hideEvent(QHideEvent *event) has been completed, we can be sure that nothing is visible any more. I don't believe it yet, but I will try it. So I will try to emit a signal at the end of hideEvent.
The cursor issue I also saw. I'll try to find out how to improve this. The current cursor mode during screen capture is not so correct at the moment.
Regarding the actions that are used in the desktop palette and in the board palette: It may be error prone in the future to use the same symbols. Because someone may decide to have the board palette on one screen and the desktop mode with the deskop palette on another screen. I think it is better if both palettes use different actions. The palettes even look quite differently. But I don't know if I want to make this decision or change this. Maybe a quick fix is to not updating the arrows in showEvent and hideEvent, but in UBDesktopAnnotationController::goToUniboard and somewhere, where showDesktopSignals (or whatever name is used) are received.
Updating the masking is a good reminder. I don't like this masking idea, at all. And it does not even work very well. But I see no alternative at the moment and it works good enough.
I will come with another draft in a few days or weeks.

letsfindaway:
Just some ideas I got:

- There is an example for a screenshot utility at https://doc.qt.io/qt-6.2/qtwidgets-desktop-screenshot-example.html. They are using a timer after hiding the application. Probably this is the way to go with a reasonable short default timer value of some hundred milliseconds for all those who do not have fancy fade-out effects and the option to make it longer for others?
- Instead of hiding just the toolbar: couldn't we hide the complete application during screenshot? And also just make the `UBCustomCaptureWindow` visible during selection of the area? This would also make the mask ineffective during this.

PR #465, opened 2021-03-20
Feature: Export background grid/color to PDF

letsfindaway:
Some users suggested to add options to export the background grid and/or color to PDF. This pull request adds these features.

They can be enabled by the following configuration variables:

```
[PDF]
ExportBackgroundGrid=true
ExportBackgroundColor=true
```

Both variables are set to `false` by default, which is also the current behavior.

This feature is mentioned in

- https://github.com/OpenBoard-org/OpenBoard/issues/403
- https://github.com/OpenBoard-org/OpenBoard/issues/357
- https://github.com/OpenBoard-org/OpenBoard/issues/315
- https://github.com/OpenBoard-org/OpenBoard/issues/118

With this PR these options are only available via the configuration file. It is for further discussion if and possibly where they should also be added to the settings page of the user interface.

PR #637, opened 2022-06-22
fix: restore PDF background export

letsfindaway:
The feature to export background color and grid to PDF files was not enabled with `UBExportFullPDF`.

This merge requests enables drawing the background according to the corresponding configuration variables also in this case.

See also https://github.com/OpenBoard-org/OpenBoard/pull/465#issuecomment-1161309848.

### DIFF
--- a/src/adaptors/UBExportFullPDF.cpp
+++ b/src/adaptors/UBExportFullPDF.cpp
@@ -108,20 +108,11 @@ void UBExportFullPDF::saveOverlayPdf(UBDocumentProxy* pDocumentProxy, const QStr
         for(int pageIndex = 0 ; pageIndex < pDocumentProxy->pageCount(); pageIndex++)
         {
             UBGraphicsScene* scene = UBPersistenceManager::persistenceManager()->loadDocumentScene(pDocumentProxy, pageIndex);
-            // set background to white, no grid for PDF output
+            // set background according to PDF export settings
             bool isDark = scene->isDarkBackground();
             UBPageBackground pageBackground = scene->pageBackground();
 
             bool exportDark = isDark && UBSettings::settings()->exportBackgroundColor->get().toBool();
-
-            if (UBSettings::settings()->exportBackgroundGrid->get().toBool())
-            {
-                scene->setBackground(exportDark, pageBackground);
-            }
-            else
-            {
-                scene->setBackground(exportDark, UBPageBackground::plain);
-            }
 
             // set high res rendering
             scene->setRenderingQuality(UBItem::RenderingQualityHigh, UBItem::CacheNotAllowed);
@@ -141,8 +132,22 @@ void UBExportFullPDF::saveOverlayPdf(UBDocumentProxy* pDocumentProxy, const QStr
 
             if (pageIndex != 0) pdfPrinter.newPage();
 
+            // do not draw background color and grid if scene has PDF background
+            if (mHasPDFBackgrounds)
+            {
+                scene->setDrawingMode(true);
+                scene->setBackground(false, UBPageBackground::plain);
+            }
+            else if (UBSettings::settings()->exportBackgroundGrid->get().toBool())
+            {
+                scene->setBackground(exportDark, pageBackground);
+            }
+            else
+            {
+                scene->setBackground(exportDark, UBPageBackground::plain);
+            }
+
             //render to PDF
-            scene->setDrawingMode(true);
             scene->render(pdfPainter, QRectF(), scene->normalizedSceneRect());
 
             //restore screen rendering quality


letsfindaway:
## Background

Some background on this simple PR. 
In OpenBoard 1.6.2, the pdf-merger was deactivated. All PDF exports were performed by `UBExportPDF`. Here the exporting of background colors and grid was correctly implemented and worked, when the corresponding configuration variables were set.

In OpenBoard 1.6.3 the pdf-merger was reactivated. Most PDF exports are now performed by `UBExportFullPDF`. Here the evaluation of the configuration variables was also present, but a small bug remained, which is explained in the following.

The original code for rendering is
https://github.com/OpenBoard-org/OpenBoard/blob/d4b91d8d34744ff33ca4d16e02dd60041cd82c52/src/adaptors/UBExportFullPDF.cpp#L144-L153

In the beginning, `scene->setDrawingMode(true);` is called. After rendering, `scene->setDrawingMode(false);` is called. The original intend of this code is to suppress drawing of the background. Let's see what this call does.
https://github.com/OpenBoard-org/OpenBoard/blob/d4b91d8d34744ff33ca4d16e02dd60041cd82c52/src/domain/UBGraphicsScene.cpp#L1181-L1184
It just sets the member variable `mIsDesktopMode`. The only use of this variable is here:
https://github.com/OpenBoard-org/OpenBoard/blob/d4b91d8d34744ff33ca4d16e02dd60041cd82c52/src/domain/UBGraphicsScene.cpp#L2693-L2699
So it just suppresses drawing of the background. As the name of the variable suggests, this was meant for desktop mode, and is just "misused" here for PDF export. But anyway, this suppression inhibits the export of background grid and color.

The patch just removes this and activates the interpretation of the configuration variables to control the background export behavior.

kaamui:
I just tried to comment the lines you removed and the pdf background is no longer present in the output pdf.

letsfindaway:
> I just tried to comment the lines you removed and the pdf background is no longer present in the output pdf.
> 
> Moreover, souldn't `ExportBackgroundColor` and `ExportBackgroundGrid` be checked somewhere ? It feels like we're going to perform these drawings (`painter->fillRect (rect, QBrush (QColor (Qt::black)));`, `painter->drawLine (rect.x (), yPos, rect.x () + rect.width (), yPos);`) even with these parameters set to `false`, with this maybe too simple PR.

The variables are checked. See the complete snippet:
https://github.com/OpenBoard-org/OpenBoard/blob/0a3cbdced44c565cedff15cc96e88b2b422e36da/src/adaptors/UBExportFullPDF.cpp#L112-L154
This is code from the `dev` branch.

kaamui:
I know that, but after, `scene->render` (line 146) is called, `drawBackground` is called by it and the portion of code mentioned is executed, regardless of these parameters' values.

kaamui:
Note that the exported PDF I obtain with your modification seems to behave as if there was no PDF on the background, but still has the same weight as if the PDF background was indeed part of the exported document. So it looks like there is the PDF on the background, another background over it (probably drawn by `painter->fillRect` and so on, as it is called no matter what), then the annotations. 

Reading you, I guess you didn't encounter the issue at all, so maybe is it OS-specific (and that for you the PDF page is rendered above the background), and somehow related to the Qt version, or the pdf library version, I don't know...

Here's what is looks like for me (with a page you should recognize) : 

image

letsfindaway:
Now I got your point. Sorry that it took so long, but I just missed that you mean the an _imported_ PDF as background. I was only thinking of the color and lines in an _exported_ PDF.

Yes, you're right. I will certainly fix that. My idea is now:

- When we have an imported PDF as background, then any handling of background color and grid is useless. Just take the PDF as background.
- When we don't have an imported PDF, then background color and grid is relevant and will be exported according to the configuration variables.

I will also double-check that this works with `UBExportFullPDF` as well as `UBExportPDF`.

letsfindaway:
The commit just pushed does the following when using `UBExportFullPDF`:

- If the page has no PDF background image, then the background color and grid is exported according to the settings.
- If the page has a PDF background, then background color and grid are not exported. The drawing colors are exported as if the background color would be white.

The reasoning behind this is as follows:

If a scene is rendered on the board, then there ate actually three layers:

- Background color and grid at the bottom.
- An optional PDF background image on to of that.
- The other elements of the scenes on top.

Background color and grid are visible outside of the area of the background PDF.

When exporting with `UBExportFullPFD` we can however only have two layers: The PDF background image at the bottom and everything else above. Therefore we cannot export a background color/grid if a PDF background image is used. It was my original error not to observe that.

If somebody is using a black background together with a PDF background image and draws annotations using the white pen outside of the PDF background area, then these annotations would not be visible in the exported document. I therefore have chosen to switch the scene's background to white during this operation, so that such annotations are becoming black. This has however another price: maybe the user did this by intent. E.g. the background PDF itself is very dark, and the user therefore switched the background color to get a white pen, then this would not be honored. So for me the question remains how to cope with this:

- Shall we leave the background color unchanged and accept that annotations outside of the background PDF image might not be visible?
- Shall we switch the background color to white to switch the pen colors and accept that the colors of annotations on the background image are different than on the board?

kaamui:
> Shall we leave the background color unchanged and accept that annotations outside of the background PDF image might not be visible?

Nope. 

> Shall we switch the background color to white to switch the pen colors and accept that the colors of annotations on the background image are different than on the board?

Yep. It's the initial behavior of OpenBoard.

Common usage (at least for teachers in Geneva) is to scan a book (or part of), to import it as PDF background in OpenBoard and make annotations above it during class. These books and their scans are most likely to have a clear background for obvious reasons. 

The use case where the PDF background is black is possible but way less likely to happen. And what to do with annotations that would start outside the PDF background and finish inside it ? 

As you stated, `exportBackgroundGrid` and `exportBackgroundColor` are of no interest if a PDF background has been imported. They should behave perfectly with no PDF Background, and be ignored with one.

However, I imagine it remains possible to achieve what you initially thought, at least without the use of pdf-merger (so you would have the grid as background, and the pdf treated as a common graphics item, placed under any other object).

Issue #638, opened 2022-06-23
Correct installation of files according to FHS


neurofreak-alt:
Greetings!
I would like to see the structure of the project installed in the %buildroot brought closer to the FHS ( https://wiki.linuxfoundation.org/lsb/fhs) used in linux. Right now everything is built and installed in the ../product/release directory. I would like to see the configuration files in ../build/release/etc/OpenBoard, binary files in ..build/release/usr/bin, resources in ../build/release/usr/share/OpenBoard Will this happen in the future?

letsfindaway:
Yes, I would also like to see that.

Currently in some places a set of patches is used to bring OpenBoard closer to the FHS layout. I'm also using patches for my [openSUSE builds of OpenBoard](https://build.opensuse.org/package/show/Education/OpenBoard). i think especially patch 0002-Install-to-correct-directories-on-linux.patch and 0004-Use-QStandardPaths-to-locate-resources.patch are important. I don't know exactly who started to create these patches, but might be they originate from the [Flathub builds](https://github.com/flathub/ch.openboard.OpenBoard/tree/master/patches).

The resulting layout is still not perfect, but at least

- the binary goes to `/usr/bin/openboard`,
- the library and all other resources files go to `/usr/share/OpenBoard`.

`/usr/share/OpenBoard/etc` is still a mess which would need some cleanup, but it is better than installing to `/opt`.

Note however that the deployment of OpenBoard at the schools in Geneva might have additional and contradicting requirements. I don't know, but it might be that it wouldn't be so easy for them to move to another directory layout. So this needs probably to be discussed and a migration strategy is needed.

JBBgameich:
The patches used for flathub or just a `git format-patch upstream/1.7-dev` from [my repository on GitLab](https://gitlab.com/JBBgameich/OpenBoard/-/commits/1.7-dev), where I'm maintaining the patches for the flatpak packaging for each branch that I'm interested in in git.
If you want we can use that (or a new repository) and maintain these patches together.

letsfindaway:
> The patches used for flathub or just a `git format-patch upstream/1.7-dev` from [my repository on GitLab](https://gitlab.com/JBBgameich/OpenBoard/-/commits/1.7-dev), where I'm maintaining the patches for the flatpak packaging for each branch that I'm interested in in git. If you want we can use that (or a new repository) and maintain these patches together.
> 
> EDIT: To be exact, the 1.7-dev branch is what the next flatpak release will be based on.

[...]

PR #641, opened 2022-06-24
Fix font handling

### LINKED ISSUE #474: Font list shows very few fonts (Text tool, OpenBoard 1.6.1, Linux)
Since last update (1.6.1) the font dialog shows very few fonts:

- deb package (openboard_ubuntu_20.04_1.6.1_amd64.deb): it shows "Andika Basic", "BSTGreek", "Ecolier_court", "Ecolier_lignes_court", "Ecriture A", "Ecriture B", "GE Typo Libre". Those fonts are not installed on my system.
- flatpak ("ch.openboard.OpenBoard"): it shows "Z003 [UKWN]". This one is installed on my system.

Running both versions from the command line (`openboard` and `flatpak run ch.openboard.OpenBoard`) doesn't show any font-related messages.

Operating system: Linux (Kubuntu 20.04.2 LTS)

letsfindaway:
This patch fixes issues with font selection for the text tool:

- Huge performance improvement when building the fonts list by
  - removing fonts from the models in chunks instead individually. This is important, as removing a single item from the model using `removeRow` takes nearly the same time as removing many items using `removeRows`. On my system with about 750 system fonts installed this reduced the time needed to build the fonts list from about 30 seconds to some 100 ms,
  - removing duplicates from the custom fonts list. This speeds-up comparison.
- Prefix matching to find all fonts.
  - The custom font list contains a prefix only, e.g. "Ecriture A", while the `dialogFontNames` contain e.g. "Ecriture A", "Ecriture A Ligne", "Ecriture A Orne" and "Ecriture A Orne Ligne". The check for equality dropped all but the first from this list.

Note that it is still necessary to add the fonts to the `INSTALLS` list in `OpenBoard.pro`. I think currently the fonts are missing in nearly all community builds, including mine for openSUSE. For the time being I will do this in a patch in my builds (see [0008-install-fonts.patch](https://build.opensuse.org/package/view_file/home:letsfindaway:experimental/OpenBoard/0008-install-fonts.patch?expand=1)), but in the end this should be addressed in a common cleanup of `OpenBoard.pro`, including considering #638. This patch will also be important for the Flatpack version maintained by @JBBgameich and necessary to resolve #474 for this version.

Additional remark: OpenBoard has font files in the directories

- `resources/fonts` and
- `resources/customization/fonts`.

Only the latter is used for the text tool. The first directory is addressed in the source code of `XPDFRenderer`, but in fact these files are not packaged in the official Ubuntu builds, so it is questionable whether they serve any actual purpose.

### DIFF
--- a/src/domain/UBGraphicsTextItemDelegate.cpp
+++ b/src/domain/UBGraphicsTextItemDelegate.cpp
@@ -263,12 +263,24 @@ void UBGraphicsTextItemDelegate::customize(QFontDialog &fontDialog)
 
             QStringList customFontList =  UBResources::resources()->customFontList();
             int index = 0;
-            foreach (QString dialogFontName, dialogFontNames){
-                if (safeWebFontNames.contains(dialogFontName, Qt::CaseInsensitive) || customFontList.contains(dialogFontName, Qt::CaseSensitive))
+            int remove = 0;
+
+            for (const QString& dialogFontName : dialogFontNames)
+            {
+                if (UBStringUtils::containsPrefix(safeWebFontNames, dialogFontName, Qt::CaseInsensitive) ||
+                        UBStringUtils::containsPrefix(customFontList, dialogFontName, Qt::CaseSensitive))
+                {
+                    stringListModel->removeRows(index, remove);
+                    remove = 0;
                     index++;
+                }
                 else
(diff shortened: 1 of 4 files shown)


JBBgameich:
Thank you, I'll apply your patch on the flatpak side too.

PR #645, opened 2022-06-25
Text fits in field ( Preferences->About )

ThomasLucky13:
Text in field of Preferences->About didn`t fit. I fixed it.

изображение

изображение

### DIFF
--- a/resources/forms/preferences.ui
+++ b/resources/forms/preferences.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>825</width>
-    <height>830</height>
+    <width>1135</width>
+    <height>958</height>
    </rect>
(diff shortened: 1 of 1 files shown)


habruening:
With the size of 1135x958 the preference dialog looks really awful. I would not accept this pull request. We should rather try to find out why the text does not fit for you or make the default size dynamic. On my configuration the text consumes only half of the size. I will try to increase the dpi number. But I think, making it always such big is a bad idea.

habruening:
What about replacing the text "Check if Open-Sangoré data could be importet at lauch" by "Check at program start if data can be imported"
And "Check software update at launch" by "Check at program start for software updates"
?

ThomasLucky13:
Yes. Now it fits.
изображение

PR #646, opened 2022-06-25
1.7 dev desktop palette hide event

habruening:
A small refactoring activity:

UBDesktopPalette::showEvent and hideEvent should not have a side effect to the palette of the main drawing board. It is relatively unexpected when going through the code that the icons of the main drawing board change magically. So that should be a bit more clear. That is the purpose of this pull request. The method hideEvent is then not even needed any more.

Note: The procedures setPenMarkerEraserIconsWithArrows and setPenMarkerEraserIconsWithoutArrows are no member functions because there is no need of beeing member functions. Everything is defined statically. In principle it should be a static member of UBApplication because there the actions are defined or even a helper outside that class. But the responsibility of this mechanism is in UBDesktopAnnotationController. So it makes sense to group all what belongs to it there.

### DIFF
--- a/src/desktop/UBDesktopAnnotationController.cpp
+++ b/src/desktop/UBDesktopAnnotationController.cpp
@@ -183,22 +183,11 @@ UBDesktopAnnotationController::~UBDesktopAnnotationController()
     delete mTransparentDrawingView;
 }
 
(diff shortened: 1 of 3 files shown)


habruening:
I will come with a bigger pull request.

PR #649, opened 2022-07-01
fix: scaling of background PDF during export

letsfindaway:
This PR tries to fix a scaling issue during PDF export using the PDF merger, where the overlay and the base PDF document where not correctly aligned. See https://github.com/letsfindaway/OpenBoard/issues/93 for an analysis of this problem.

I tested the patch also for cases where the annotations cover areas outside of the PDF document size. In these cases the PDF document and the annotations are scaled, so that everything fits on the page. Still the annotations are at the correct place.


letsfindaway:
The PR is now rebased to `dev` and written in a way that it is compatible with the reworked `UBDisplayManager` in the `1.7-dev` branch.

PR #651, opened 2022-07-07
chore: reorganize linux build

letsfindaway:
This pull request aims to reorganize the Linux builds to enable a more FHS conformant layout, while keeping the option to create a package structure very close to the currently released official packages. Builds for Windows and OS X should not be affected.

To enable this, the project file uses the following environment variables to control file system layout:

| Variable | Default | Remark |
| --- | --- | --- |
| `PREFIX` | `/usr` | Common prefix for the install location |
| `SYS_PREFIX` | `$PREFIX` | Prefix for system files: binaries, desktop file, icons and mime type specification |
| `APP_PREFIX` | `$PREFIX/share/openboard` | Prefix for the application data files: default settings, library, languages, fonts. |

The default build now creates a largely FHS conformant file system layout. The binary goes to `/usr/bin`, the application files are in `/usr/share/openboard`, desktop file in `/usr/share/applications`, icons in `/usr/share/icons` and mime specification in `/usr/share/mime`.

By setting `APP_PREFIX=/opt/openboard` a layout is created, which is quite similar to the current package layout. Differences are:

- `/usr/bin/openboard` is not a symlink, but the actual executable.
- Icons and mime specification are installed in the standard system locations.

The latter makes `postinst`and `prerm` scripts obsolete, as the package installer automatically registers these files and updates the system databases.

The project file now installs all necessary components including fonts, files in `/etc`, desktop file, icons, mime specification in their respective directories, making it unnecessary to copy these things to the package in the `package.sh` script. The installation directory now contains everything what is needed to create a valid package.

Locating libraries now makes use of the `pkg-config` infrastructure, so that no absolute paths are necessary in the project file. The obsolete `UB_THIRDPARTY_INTERACTIVE` was removed.

The desktop file is now part of the repository. Icon files and mime specification files start with a `ch.openboard` prefix, which serves as app id for Flatpack builds. A scalable SVG icon is now also used for the application.

The `build.sh` script has only minor modification. For the `package.sh` script however more changes have been necessary.

- It is no longer necessary to run this script as `root`, so the corresponding check was removed.
- As the created installation directory contains all necessary files, it is no longer necessary to copy them.
- It is no longer necessary to create a desktop file and a `postinst` script.

The pull request also contains a small improvement: When the `onboard` on-screen keyboard is not installed, the corresponding option is forced to use the built-in keyboard. To show and hide the system `onboard` on-screen keyboard, the appropriate DBus messages are used instead of starting and killing the `onboard` process each time. These will also automatically startup `onboard` once, if it is not running. See https://develop.kde.org/docs/use/d-bus/dbus_autostart_services/ for description of this mechanism.

Many ideas are based on the work of @JBBgameich - thanks for this and also for the discussion at https://github.com/letsfindaway/OpenBoard/issues/96. I hope that this PR is also a big step forward for Issue #638. It should also ease community builds.


JBBgameich:
Thank you for making this flatpak-ready, this will save me a lot of work in the future :)

Btw, @kaamui, would you be interested in me upstreaming the appstream metadata? That file contains the changelog and some other app store metadata which is displayed for example in GNOME software and Plasma Discover. But it needs to be updated on each release.

letsfindaway:
The `package.sh` script is now also adapted for the case that the Qt libraries are bundled with the package. The former `run.sh` script is now replaced by a dynamically created script which is installed as `/usr/local/bin/openboard`. In Ubuntu and most other distributions `/usr/local/bin` is in the standard `PATH` before `/usr/bin`. Therefore the same application desktop file can be used, which is now invoking this script. The script sets the required environment variables and launches `/usr/bin/openboard`.

The mechanism was tested on a fresh installation of Ubuntu 20.04 LTS and a bundled Qt 5.15.2.

letsfindaway:
Superseded by #830

PR #652, opened 2022-07-08
Fixed the program version in the text from the copyrightTextBrowser widget

neurofreak-alt:
Please accept a small correction. The problem was that the version in the "About" help text does not match the version in the package name and the version listed at the top of the program settings window.
OpenBoard-about-versions-conflict


kaamui:
thank your for this pull request, but actually I was going to remove it so it is no longer needed to update it.

PR #668, opened 2022-09-21
Portable version

SeeDborg:
Hello everyone. 
It would be really cool to have a portable version of OpenBoard (for example to put on a USB key, for teachers who go from school to school...).
Regards !


SeeDborg:
Is it something that will be available in the upcoming 1.7 version ?
I have many demands about this from teachers :)

kaamui:
on what OS ? If on Windows, it is already possible, you won't have the register keys (registering the icons for ubz and ubx documents if I remember well), just copy the installation folder to your USB key. But you would not have your documents copied here. You can maybe try to change the configuration file by adding the `DataDirectory=path/to/your/key` but I'm not even sure it could work with a different drive (never tested). And I don't think the letter of your drive is ever guaranteed so...

The best thing to do in my opinion, is to install openboard normally on each school, and to export "My Documents" as UBX to your USB key each time you move from one school, so you can then import it in the other one.

SeeDborg:
I think that the idea of having a fully portable version of OpenBoard would be a great adding (I download it, I put it on my usb key, I use it).

Alternatively, is it possible to put a relative path for the DataDirectory ? (like ./Documents, so we can just create a "Documents" folder inside the installation folder of Open board).

Anyway, is there a particular reason to store the Documents in the user folder by default ? Why not just store it where OpenBoard is installed ? I guess this is to have one document folder per user, but is it really the most use scenario ?

kaamui:
> Alternatively, is it possible to put a relative path for the DataDirectory ? (like ./Documents, so we can just create a "Documents" folder inside the installation folder of Open board).

I don't know and can't look at it. Try and let me know.

> Anyway, is there a particular reason to store the Documents in the user folder by default ? Why not just store it where OpenBoard is installed ? I guess this is to have one document folder per user, but is it really the most use scenario ?

You answered your question yourself, it would be disastrous for many reasons.

SeeDborg:
> I don't know and can't look at it. Try and let me know.
I'll try and tell you :)

> You answered your question yourself, it would be disastrous for many reasons.
I don't think that anyone is using OpenBoard for multiple users. In most cases, there is one computer in each classroom, and one user (teacher) using it.

SeeDborg:
It seems to work !
If I put DataDirectory=./documents I have an error, BUT
DataDirectory=./library/documents seems to work !
I'll continue to test and keep you updated :)

SeeDborg:
But once more, I find that the option to change the directory should be in the apps settings and not in an obscure config file.
(forgive me if I'm insistant but I really love your software and try to propose solutions so it'll be even better ! I hope I'm not too annyoing ^^)

kaamui:
Don't worry^^

It will probably take a while before I can look at it, or it may not be validated internally, but I intent to add two tabs in the preferences : one listing the changes of each new version, and another one listing all the configurable options.

SeeDborg:
Cool !
When you release 1.7, I can, if you want, make a portable version using the trick I described, so you can add the "portable" version to the download options.
Because once more, I think that many teachers won't be able to use that on their own.

SeeDborg:
A little question for the future though : when you have an update, do you keep the settings in the .config file, or do you reset them to the default values ?
I don't want to create any bugs.

kaamui:
A new version always come with its own config file (potentially with some options removed, some others added). 

I don't want to promote a portable version as it is not cross-platform, and would work with some "hard-coded" configurations that I don't want to encourage, but thank you, and the most important is that you can on your side adapt OpenBoard to your needs.

SeeDborg:
OK, I hope to see an official portable version someday then :)

PR #669, opened 2022-09-22
Fix memory leak in UBSvgSubsetAdaptor

letsfindaway:
This PR fixes two memory leaks in `UBSvgSubsetAdaptor`. 

- In `UBSvgSubsetWriter::strokeToSvgPolygon` a temporary object is never deleted.
- In `UBSvgSubsetReader::pixmapItemFromSvg` an object would not be freed in an error case.

Additionally it avoids several potential `nullptr` dereferences by moving the affected code in a guarding `if`. A third commit removes two lines of code where a `QString` is constructed which is never used later.


letsfindaway:
Converted to Draft as I'm adding more such things to this PR.

PR #672, opened 2022-09-26
fix: correct calculation of graphics item margins

letsfindaway:
- boundingRect only adds a margin if
  - it is a QAbstractGraphicsShapeItem, and
  - it has a pen (style != Qt::NoPen)
- the margin is then half the pen width at each side


letsfindaway:
Have to double-check this with the PDF export. I now observe a 0.5 pix offset between annotations and PDF background, which was not here before.

letsfindaway:
Sorry, false positive. All right with this PR.

kaamui:
Thank you for your reactivity !

PR #676, opened 2022-09-30
fix: create url from local file when freezing widget

### LINKED ISSUE #675: Export with WebWidget failed (V1.6.3 Windows)
Hello,
When I export document in PDF, WebWidgets application (As GoogleMap or OpenStreetMap) presents an QTNetwork Error 301 : "Failed to load URL c:/Program Files/OpenBoard/etc/freezedWidgetWrapper.html." Also, the file freezedWidgetWrapper.html exists.
Any idea why ?

Regards,
Olivier

Capture d’écran 2022-09-29 194457

letsfindaway:
When freezing a widget in `UBBoardController::freezeW3CWidgets` the board controller passes the file name to the `QUrl` constructor. On Windows, this causes an "Unknown protocol" error, because the drive letter is interpreted as protocol. Use `QUrl::fromLocalFile` instead.

On Linux and MacOS there is no drive letter and the browser engine correctly interprets the passed file name.

This problem is only visible during PDF export, because the frozen widgets are on inactive pages only, where the error message is not visible to the user.

Fixes #675


obook:
The problem is same under Linux, but with black frames.
Screenshot_20221002_124607

letsfindaway:
Closed as #680 was merged.

PR #681, opened 2022-10-08
feat: open ubz file in running instance

letsfindaway:
This PR fixes the "known issue" and allows to double-click on an ubz file while OpenBoard is running. It will

- send a message with file name from the secondary instance and exit,
- handle message in primary instance and open the file.

This was already partly implemented, just the open ends have to be glued together to make this working.

### DIFF
--- a/src/core/UBApplication.cpp
+++ b/src/core/UBApplication.cpp
@@ -91,7 +91,7 @@ bool bIsMinimized = false;
 QObject* UBApplication::staticMemoryCleaner = 0;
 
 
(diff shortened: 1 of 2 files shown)


kaamui:
Stop doing my work ! :smile: 

(I had started to look at singleApplication's github to look how to do it, with secondaryInstance etc.. as it is one of the issues the team wanted to prioritize)

Just kidding, thank you for this. I didn't see that we already had a similar function as the "message receiver" with UBApplicationMessage::handleOpenMessage.

letsfindaway:
> Stop doing my work ! :smile:

Not sure whether I can promise that ;)

PR #685, opened 2022-10-12
Check widget compatibility

letsfindaway:
This PR improves the handling of widget compatibility when upgrading from OpenBoard 1.6.x to 1.7.x.

- It adds an explicit API version marker to the `config.xml` file, which identifies widgets compatible with the asynchronous API.
- It shows a note on the screen for incompatible widgets.
- It includes the location of user installed widgets (`~/.local/share/OpenBoard/interactive content/` on Linux, similar locations on other platforms) when searching for upgraded widgets.

The version marker is already documented [in my Wiki page](https://github.com/letsfindaway/OpenBoard/wiki/Widget-API#widget-upgrade).

### DIFF
--- a/resources/etc/widgets.md5sum
+++ b/resources/etc/widgets.md5sum
@@ -1,3 +1,5 @@
+2538bef70e75fcc6c3ebe2ec35e0a904,http://www.openboard.org/widget/application/geoinfo,GeoInfo.wgt
+7ade330c8afe00e543169486f65b7ccf,http://www.openboard.org/widget/application/icell,iCell.wgt
 0e417c056f2725a7147c12a74cc17bfd,http://www.openboard.org/widget/interactivity/assaudio,Ass audio.wgt
 183b3cee755f695eef5fa7293d0acd32,http://www.openboard.org/widget/interactivity/assimages,Ass images.wgt
 dadbfcf9fe185fcb3aa15cacec3d1aa1,http://www.openboard.org/widget/interactivity/calculation,Calculation.wgt
--- a/resources/library/applications/GeoInfo.wgt/config.xml
+++ b/resources/library/applications/GeoInfo.wgt/config.xml
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <widget xmlns="http://www.w3.org/ns/widgets" 
+        xmlns:ub="http://uniboard.mnemis.com/widgets"
+        id="http://www.openboard.org/widget/application/geoinfo"
         version="2.0"
         width="680"
         height="400" 
-		ub:resizable="false">
-  <name>GeoInfo</name>
-  <content src="GeoInfo.html"/>
+        ub:resizable="false">
+    <name>GeoInfo</name>
+    <content src="GeoInfo.html"/>
 </widget>
--- a/resources/library/applications/iCell.wgt/config.xml
+++ b/resources/library/applications/iCell.wgt/config.xml
@@ -1,12 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <widget xmlns="http://www.w3.org/ns/widgets" 
-        identifier="http://uniboard.mnemis.com/widgets/Clock" 
-        version="1.1"
+        xmlns:ub="http://uniboard.mnemis.com/widgets"
+        id="http://www.openboard.org/widget/application/icell"
+        version="2.0"
         width="500"
         height="500" >        
-  <name>iCell</name>
-  <content src="widget.html"/> 
+    <name>iCell</name>
+    <content src="widget.html"/>
 </widget>
 
 <!-- This XML is probably more complete, but it isn't working very well (the widget is resized, but its content doesn't move)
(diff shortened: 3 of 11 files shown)


kaamui:
These commits will make custom widgets treated as incompatible because their config.xml does not contain the mention `ub:apiVersion=2`, regardless of them actually using the api or not, and thus regardless of them working or not. I think it's hard to determine if a widget is really incompatible, unless it explicitly uses the API, but this also could be hard to determine. 

It makes me wondering if having this dedicated version attribute for the API is really needed. On a side note, if it is, the attribute should also be added to `UBEmbedController::generateConfig`

What do you think ?

letsfindaway:
> These commits will make custom widgets treated as incompatible because their config.xml does not contain the mention `ub:apiVersion=2`, regardless of them actually using the api or not, and thus regardless of them working or not. I think it's hard to determine if a widget is really incompatible, unless it explicitly uses the API, but this also could be hard to determine.

Also custom widgets need to be aware of the API version, if they use it. So if - just as an example - Papier would still be a custom widget, then the author would have to care about adapting it. And then the author should also add the API version marker.

> It makes me wondering if having this dedicated version attribute for the API is really needed. 

I started without that marker and was using the widget version for that. This turned out to be impracticable, because you cannot make any assumptions about the usage of that version field.

> On a side note, if it is, the attribute should also be added in `UBEmbedController::generateConfig`

Very good point. Indeed, these widgets generated by the `UBEmbedController` never use the API and never have to be upgraded.

> What do you think ?

I think I have to think twice ;) Indeed I missed, or at least neglected the case of widgets which do not use the API at all. And all widgets generated by the `UBEmbedController` are such widgets.

The easiest step would be just to remove the compatibility check and just load and try widgets without API version marker and without an update candidate. If it just works - fine! If not - I just fear that users then complain about OpenBoard, and not about the widget. That was the reason why I thought an explicit message would be helpful.

letsfindaway:
The new API is not in all cases incompatible. This is only the case if

- it uses a **function**, which is now asynchronous,
- it uses a **signal**, which is now also asynchronous,
- it uses a **slot** and does not exactly observe the parameter types.

We probably have to distinguish the following cases:

- Widgets which do not use the API at all, like those created from the `UBEmbedController` or previously `UBTrapFlashController`.
- Widgets which only use compatible parts of the API, like `Horologe`.
- Widgets which don't observe the parameter types and can be fixed in a compatible was, like `Papier`. After fixing it would belong to the previous category.
- Widgets which use functions and signals and must be adapted.

I wrote a shell script using `grep` which can distinguish these groups except for the third (`Papier`). Such things can only be found by testing. But I assume we don't want to search through all the widget code each time a document is loaded to determine API usage. Therefore I still think an API version marker is useful, because it eases the task of the `UBWidgetUpgradeAdaptor` very much: if it detects a widget with such a marker, then it can stop searching for an upgrade.

It can however be an option to implement such a scanning in C++ code and then store the result as a marker in the `config.xml`. Then this time-consuming operation would only be needed once when opening a document in the new OpenBoard version. results of the scanning are, as explained above:

- **No API**. Is compatible, because it does not use the API at all.<br>`ub:apiVersion="0"`
- **Compatible API**. Is compatible, because it only uses compatible parts of the API.<br>`ub:apiVersion=">=1"`
- **Incompatible API**. Needs upgrade, because it uses a `function` or a `signal`, which is now asynchronous.<br>`ub:apiVersion="<=1"`
- **New API**. Compatible with 1.7.x, but will not run on 1.6.x, because it uses the asynchronous API.<br>`ub:apiVersion=">=2"`

Only for the third case we would look for a new version of the widget. If we don't find one, then we should show a note to the user.

kaamui:
> > What do you think ?
> 
> I think I have to think twice ;) 

Think twice, code once ! A rule I'm forgetting these days...

> It can however be an option to implement such a scanning in C++ code 

It would be a better solution indeed, because portable. I found this example in Qt that probably contains some of the logic we need to do so => https://doc.qt.io/qt-5/qtwidgets-dialogs-findfiles-example.html

letsfindaway:
> > Then this time-consuming ...
> 
> Do you think it would be noticeable ? 

You're right, not really. I implemented a scanner going over `.js` and `.html` files and skipping some well-known like `jquery` and as an example let it scan all 43 applications and interactivities and it took all together 44ms.

The scanner however missed `Papier`. It uses only compatible API calls, but it has a type mismatch when calling `sankore.addObject()`, which fails with `QWebEngine`. This cannot easily be detected by scanning. So we would have to implement another test.

As you might know, I have to use the MD5 sum of `config.xml` to detect some of the widgets, because they do not have a unique id. This is also the case for `Papier`. So I could just say: If I needed the MD5 sum, then this is also a reason to upgrade the widget to fix the id. 
> 
> > ub:apiVersion=">=1"
> 
> Also, do we really need to keep the information that it is compatible in the "noApi" way or "new API" way ? Because I think that in the end we're just going to do nothing, or trying to upgrade the widget, so maybe something like `widgetCompatibilty=true` and `widgetCompatibilty=false` is enough ?

Understood. And yes, I now can see that this is a feasible way. Initially I thought scanning would be harder work. But the `QRegularExpression` based scanners are easy to write and are highly optimized.

The procedure would then be as follows:

- Scan widget for compatibility.
- (only if compatible) Identify widget by MD5 sum or id.
- If (incompatible OR MD5 sum was needed for identification) AND we have that widget in our library, then upgrade.

Issue #591, opened 2022-04-18
 [Feature Request] Move through the document while holding space


SeeDborg:
I think all is in the title :)
It would be a nice feature, it's present in many softwares nowadays.

SeeDborg:
Wow, I see that Openboard 1.6.4 add nice features to navigate and zoom through a document. So I up this topic since I think it's related and it would help a lot to navigate.

letsfindaway:
> I think all is in the title :)

Actually, for me not really. Do you mean you want to switch to the next page by pressing the space bar? Or do you have some other action in mind?

If you want to switch pages by using the keyboard, then you can use the left and right arrow keys.

SeeDborg:
No, I don't mean moving from one page to another, but moving through the current page while I hold space and clic at the same time. Like in pdf or in photoshop. Here is an example :
![Animation](https://user-images.githubusercontent.com/7523821/194781097-92abdf9c-b1ad-4d20-bfc8-cb6fa9fb56d4.gif)

kaamui:
It's already possible with the "hand" tool

SeeDborg:
No.
What I describe is : 
 - Maintaining space bar is switching to the hand tool as long as I hold it.
 - Releasing the space bar is going back to the previous tool.

You can try it in any version of adobe reader, adobe photoshop, and many other softwares.
I think the feature is named something like : hold space to drag and pan.

SeeDborg:
A similar feature that has just been added to OpenBoard would be : https://github.com/OpenBoard-org/OpenBoard/pull/343#event-7569129967

kaamui:
OK. It would be indeed a nice feature, in terms of user experience. I'l try to ask the team if we add it to the todo list, but be aware that the OpenBoard's team is very small and that such a feature won't be a priority at all, for now. 

If you are willing to try to develop it, please provide a pull request and I would be glad to accept it.

SeeDborg:
Thanks.
Sadly I cannot help, I know my way in php but not in C++ :(

SeeDborg:
Another idea : same option but with middle mouse click :)

PR #460, opened 2021-03-14
Shortcut configuration

letsfindaway:
This pull request implements a flexible configuration of shortcuts for actions triggered by keyboard, special mouse buttons or tablet stylus buttons. It is still draft, as intensive testing is still needed. Here a description of what was added or changed:

- The preferences have a new Shortcut tab showing a list of actions and associated shortcuts.
- The actions are grouped. Grouping is hard-coded in the `UBShortcutManager` and can be extended there.
- The list can be filtered by typing any part of a description or shortcut.
- When an action is selected, the shortcuts for that action can be edited by "recording". This means, that you press the "Record" button and then any key combination, mouse button and/or stylus button you want to assign to that action. When finished, press "Accept" or "Abort". When accepted, the shortcut(s) are assigned to that action.
- You can reset the shortcuts to the application defaults using the "Reset" button.
- All mouse buttons except left and right and all stylus buttons can be used to trigger actions.
- When you type a conflicting shortcut, you will be warned and the offending entry is shown in the table.
- There is a checkbox to allow all shortcuts which are combinations using the Ctrl key also to be triggered without pressing Ctrl. If the Alt key is specified in a shortcut, this is unaffected by this checkbox.
- The configured state is persisted in the user settings.
- All shortcuts can also be triggered when the owning palette is not shown. I.e. it is no longer necessary to display the stylus palette in order to use Ctrl+I, Ctrl+E etc.

The pull request especially needs intensive testing on Windows and Mac, as I'm developing on Linux and do not have other platforms available for testing. Any feedback is highly appreciated!

My branch is based on the latest release candidate https://github.com/OpenBoard-org/OpenBoard/releases/tag/v1.6.1rc-0309, so all features and bug fixes of that release are also included.

PR #686, opened 2022-10-12
Shortcut configuration, also for Qt6

letsfindaway:
Updated pull request (was #460) to resolve merge conflicts and to enable compilation with Qt 6.

Note that this PR would also provide something similar to #591, but not with the Space key (as Space cannot be a shortcut), but with the shortcut key assigned to the Hand tool.

### DIFF
--- a/resources/forms/preferences.ui
+++ b/resources/forms/preferences.ui
@@ -1105,6 +1105,175 @@
        </item>
       </layout>
      </widget>
+     <widget class="QWidget" name="shortcutTab">
+      <attribute name="title">
+       <string>Shortcut</string>
+      </attribute>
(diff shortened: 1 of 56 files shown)


MasFlam:
I was minutes away from starting work on this myself haha. Good to see it is to be present in 1.7.0 and you have nightly builds going, so thanks a lot!

letsfindaway:
@MasFlam: Perhaps one clarification: this feature will **not** be present in the official release distributed here and on openboard.org. This PR is currently not merged and there is no fixed date for when and whether this feature will be included.

It is however part of my experimental builds (see https://github.com/letsfindaway/OpenBoard/issues/89) for openSUSE, Ubuntu and Debian. These builds are not "nightly", but I update them whenever I think it is worth doing so.

PR #687, opened 2022-10-13
Can the masking mechanism be removed?

habruening:
This pull request is only a **draft**. I raised it only because now I have got a question about the next steps.

I made many code refactorings. This reduces the complexity so that we also have 1478 less lines of code. And this without any functional change.

But now I have got a question: Most code of UBDesktopAnnotationController is the masking mechanism. But this masking mechanism does not even work well. For example if you move the desktop palette the mask (especially around the property palettes) is not updated correctly. And the selector does also not work well with the masking as it is not possible to click through drawn things. It is questionable if the masking mechanism is then useful, at all.

What about removing the masking mechanism? I have the impression, that it will never work correctly. The alternative is to have clearly two modes. One with all the drawings where it is not able to interact with the windows behind it. And one without all the drawings. It think, this will be as comfortable as now but would eliminate all these complicated masking algorithms.

### DIFF
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -487,21 +487,10 @@ QString UBBoardController::truncate(QString text, int maxWidth)
     return fontMetrics.elidedText(text, Qt::ElideRight, maxWidth);
 }
 
-
-void UBBoardController::stylusToolDoubleClicked(int tool)
-{
-    if (tool == UBStylusTool::ZoomIn || tool == UBStylusTool::ZoomOut)
-    {
-        zoomRestore();
-    }
-    else if (tool == UBStylusTool::Hand)
-    {
-        centerRestore();
-        mActiveScene->setLastCenter(QPointF(0,0));
-    }
+void UBBoardController::restoreScroll(){
+    centerRestore();
+    mActiveScene->setLastCenter(QPointF(0,0));
 }
-
-
 
 void UBBoardController::addScene()
 {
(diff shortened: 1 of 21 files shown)


habruening:
I tested OpenBoard on Windows. On Windows the Desktop Annotations and clicking through them works perfectly. So I will be careful not to break this on Windows. Windows really does not need the masking mechanism.
So I think, the masking mechanism is a bad workaround for Linux and MacOS only. For Windows it is not needed anyway. And therefore I prefer and try to remove it and replace it with something that works better. I'll continue investigating how.

kaamui:
Sorry to have to say this, but there's no chance I accept this pull request. You're making too much changes and putting them all in one PR that is supposed to concern Desktop Mode if I understood well. It would take too much time for it to be properly reviewed and tested, and I'm not sure to understand what is the gain, as I'm pretty happy (and users too) with actual Desktop Mode behavior, even if it can be improved, as anything. There are also some questionnable commits where you juste delete things because you're not sure of what they do.

So no problem that you continue this and make your own OpenBoard with your changes, but please don't spend too much energy on it if your goal is only to have it merged.

habruening:
That was the purpose of this pull request. To ask you how I could provide good contribution to Open Board. I have a few functional things that I want to bring in. But I have not started with them because the code at the moment has certain design flaws that makes it hard to maintain. This is what most projects prefer: Doing refactoring early, before implementing new features.

How do you think, I can contribute to the correction of "technical debt"?

Issue #682, opened 2022-10-09
Editable text boxes are too small (patch included)


GavinSmith0123:
I find the text box created by the text tool far too small. It should be larger by default or allow the user to specify the size by clicking and dragging.

The following change to   `board/UBBoardView.cpp (UBBoardView::mouseReleaseEvent) <UBStylusTool::Text>` means that if the user clicks and drags to create a text box, the width will be as specified. If they have not done this, the box will be larger by default.

````
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -1348,7 +1348,10 @@ void UBBoardView::mouseReleaseEvent (QMouseEvent *event)
 
                 textItem->setTextInteractionFlags(Qt::TextEditorInteraction);
                 textItem->setSelected (true);
-                textItem->setTextWidth(0);
+                if (rubberRect.width() > 100)
+                    textItem->setTextWidth(rubberRect.width());
+                else
+                    textItem->setTextWidth(1200);
                 textItem->setFocus();
             }
         }

````

This is an improvement for me, although it would be better if the 1200 was not hard coded here and depended on the current zoom level and area of the scene that was currently displayed.

sebojolais:
I have the same experience. I spent lot of time to increase the size of the TextItem each time I need the Write text.
I improved the proposed patch in 2 ways:
- the user can specify the width of the text more accurately by dragging
- else the width is set to fill the screen from the mouse point to the right border minus 5%. It works even if the zoom factor != 1.
The patch is [here](https://github.com/OpenBoard-org/OpenBoard/pull/688);

It is an improvement for me but the best behavior should be discussed.

GavinSmith0123:
Here is a video of what happens in SMART Notebook 11. You see that the text boxes expand as required as more text is entered into the boxes, until the boxes reach the right sight of the page, when the width becomes fixed. It's possible that this may be hard to achieve with Qt in OpenBoard, and if so, the patch from @sebojolais above is a big improvement in any case.

https://user-images.githubusercontent.com/3398468/197159220-15f6c8e3-a50d-476f-b8dc-bc41e195443e.mp4

sebojolais:
@GavinSmith0123 It seems great. And it should be possible in Qt. 

The default behavior of a QGraphicsTextItem is to expand with the text. Have a look at the [setTextWitdth ref](https://doc.qt.io/qt-6/qgraphicstextitem.html#setTextWidth) with width = -1.
[...]

PR #688, opened 2022-10-14
Text item is created with a larger width. If a selected area is used,…

sebojolais:
… the Text item has the width of this selected area. If there is no selected area, the width is set to fill the screen from the mouse point to the right border minus 5%. It works even if the zoom factor != 1. Link to this [issue](https://github.com/OpenBoard-org/OpenBoard/issues/682).


kaamui:
Hi,

thank you for this pull request. In its current state I can't accept it, but it's not because of your work. The way the textbox size is currently handled has to be reworked entirely. I like your idea, but I would like for the text box to behave the same way for heightm and that is is not constantly adjusted when going to line, for example. So when I start doing this rework, I'll try to introduce your improvement too because I think it is a great idea.

Issue #671, opened 2022-09-26
Duplicate files in OpenBoard document


letsfindaway:
There are two problems leaving superfluous PDF files in an OpenBoard document, which can be demonstrated in sequence.

## Steps to reproduce

1. Open your operating system's file browser.
2. Start OpenBoard in Board mode with a new or existing document and navigate to the last page.
3. Bring the file browser to the foreground, e.g. using Alt-Tab.
4. Drag a PDF document from the file browser to the board. The pages of the PDF document are correctly inserted as new pages after the current page.
5. Close OpenBoard to avoid #670 .
6. **Problem 1:** Check the OpenBoard document folder: it contains two copies of the PDF file: one in the document's root folder and one in the `objects` folder. Only the latter is referenced in the `svg` files.
7. Now reopen OpenBoard and go to Documents mode.
8. Select all pages created by the PDF import and delete them.
9. **Problem 2:** The PDF file remain in the `objects` folder even if it is no longer referenced anywhere.

This leads to accumulated trash in the `objects` folder. A good example is the document you provided to demonstrate the PDF export offset issue. It contains 29 PDF files, some of them are identical (**Problem 3**), From those only 4 files are referenced. Two of them are even identical.

## Proposals for improvement:

- [ ] Do not create additional copies of a PDF file in the document's root directory when added using DnD.
- [ ] When deleting the last page referencing an object, delete the object.
- [ ] Before adding a new object check, whether the identical file is already there and reference that instead.

letsfindaway:
closing as #689 was merged

PR #689, opened 2022-10-20
Duplicate files

letsfindaway:
This PR addresses #671 with the following changes:

- It removes the `shouldLoadFileData` check in `UBBoardController` and never copies dropped files here, because in all cases (audio, video, images, widget, pdf) the files are copied anyway later.
- It adds a `UBPersistenceManager::cleanupDocument` function which removes unreferenced files from the document.

### DIFF
--- a/src/document/UBDocumentProxy.cpp
+++ b/src/document/UBDocumentProxy.cpp
@@ -43,6 +43,7 @@ UBDocumentProxy::UBDocumentProxy()
     : mPageCount(0)
     , mPageDpi(0)
     , mPersistencePath("")
+    , mNeedsCleanup(true)
 {
     init();
 }
@@ -54,12 +55,14 @@ UBDocumentProxy::UBDocumentProxy(const UBDocumentProxy &rValue) :
     mMetaDatas = rValue.mMetaDatas;
     mIsModified = rValue.mIsModified;
     mPageCount = rValue.mPageCount;
+    mNeedsCleanup = rValue.mNeedsCleanup;
 }
 
 
 UBDocumentProxy::UBDocumentProxy(const QString& pPersistancePath)
     : mPageCount(0)
     , mPageDpi(0)
+    , mNeedsCleanup(true)
 {
     init();
     setPersistencePath(pPersistancePath);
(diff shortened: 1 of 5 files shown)


src/document/UBDocumentProxy.cpp
```diff
@@ -134,6 +137,13 @@ void UBDocumentProxy::setWidgetCompatible(const QUuid &uuid, bool compatible)
     mWidgetCompatibility[uuid] = compatible;
 }
 
+bool UBDocumentProxy::needsCleanup()
+{
+    bool tmp = mNeedsCleanup;
+    mNeedsCleanup = false;
```
kaamui, line 143: It's quite odd to do it this way. Can you change it to the "classical" getter/setter method to avoid future confusion ? 

`if (myDoc->needsCleanup())` doesn't sound like you're modifying myDoc

letsfindaway: The idea was to make sure in the function that this returns `true` only exactly once. It's like a fuse that can only blow once. If I use classical getter/setter, then this logic would have to be at the caller side. I'm trying to find other examples for that, an probably a better name. E.g. the atomic functions use something like `fetchAndStore` to modify a value and return the previous one.

kaamui: still in this case I'm not sure it would add more benefit than it reduces readability. Maybe something like  `mCleanupPerformed` 

with, for example : 
```
if (!cleanupPerformed())
{
...
// unreferenced file or directory
                if (entry.isDir())
                {
                    qDebug() << "Deleting unreferenced directory" << filename;
                    if (!UBFileSystemUtils::deleteDir(absoluteFilePath))
                             qWarning() << "something went wrong during the cleanup process on directory" << filename;
                }
                else
                {
                    qDebug() << "Deleting unreferenced file" << filename;
                    if (!UBFileSystemUtils::deleteFile(absoluteFilePath))
                             qWarning() << "something went wrong during the cleanup process on document" << filename;
                }

...
mCleanupPerformed = true;
}
```

If you still think it is important and want to maintain it as-is, please provide a name that makes it really clear

letsfindaway: Indeed I would like to stick to the semantic, as **implementing** it is part of the `UBDocumentProxy`, while **using** this function is part of `UBPersistenceManager`. My proposal is to rename it to `testAndResetCleanupNeeded` to refer to the test-and-set semantics used for atomic variables. Here the atomic behavior is not important, as the operation occurs on the single event thread, but the operation is the same.

PR #692, opened 2022-10-21
Changed the behaviour of the masking mechanism: When the selector is …

habruening:
I changed the behaviour of the masking mechanism. In the current version it does not work. If the user has added annotations over the desktop, he has no chance to work with the desktop any more. The masking mechanism prevents him from interacting with the desktop.

So I improved this behaviour. As the selector button has no use in that case (because you cannot click through annotations anyway), I changed the behaviour of the selector. It hides all annotations and minimizes the desktop palette. Then the user can operate with the desktop. When he wants the annotations back, he just maximizes the desktop palette again.

Hiding the desktop palette may be a bit surprising in this situation. But as the user cannot draw annotations, there is actually no real need of having the desktop palette maximized.

### DIFF
--- a/src/desktop/UBDesktopAnnotationController.cpp
+++ b/src/desktop/UBDesktopAnnotationController.cpp
@@ -673,7 +673,11 @@ void UBDesktopAnnotationController::selectorActionPressed()
 
 void UBDesktopAnnotationController::selectorActionReleased()
 {
+#ifdef UB_REQUIRES_MASK_UPDATE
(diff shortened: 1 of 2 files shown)


habruening:
Pull request #895 is an alternative to this.
This pull request solves the problem by not showing the annotations in selector mode, at all. Most teachers are happy with this solution.
Pull request #895 solves the same problem by making the board transparent for mouse requests. This is the better solution, but uses unusual features of Qt, that may be removed in future Qt versions. In case request #895 is accepted, this request becomes obsolete.

PR #698, opened 2022-10-28
chore: add cmake build system

letsfindaway:
This draft PR is to create awareness that I'm working on a `cmake` based build system for OpenBoard.

- add a cmake based build system which is
  - easy to understand
  - easy to maintain
  - covering Linux, Windows and MacOS platforms
  - allow to build for Qt5 and Qt6 without modifying the cmake files
  - reasonable structure in QtCreator
- create a structure with
  - concise root CMakeLists.txt
  - separate files for each src subdirectory
  - separate files for finding dependencies
- currently missing
  - support for Windows only rudimentary (WIP)
  - support for MacOS only rudimentary (WIP)
- flexible install locations
  - add `UBPlatformUtile::applicationEtcDirectory()`
  - use it whenever referencing files in this directory
  - allow to configure these directories at build time and runtime
  - create complete installation tree

The current status can be used to build OpenBoard for Linux with the following commands, starting in the checked-out source tree:

```sh
mkdir build
cd build
cmake -S .. -B . -DCMAKE_INSTALL_PREFIX:PATH=/usr -DSYSCONF_INSTALL_DIR:PATH=/etc
cmake --build .
DESTDIR=install cmake --install .
cpack -G DEB (or RPM)
```

The result is a complete, mainly FHS conforming installation tree in the `build/install` directory and an installable package in the `build` directory.

For building with a specific Qt version add `-DQT_VERSION=<5|6>` to the `cmake` command line. Only specify the major version here. When Qt is installed in a non-system directory (e.g. Qt versions downloaded from Qt and not installed from the distribution), add `-DCMAKE_PREFIX_PATH=/path/to/local/Qt/installation`.

When building on the Open Build Service, a build of a Debian-like package requires a `debian.rules` file as simple as this:

```make
#!/usr/bin/make -f
%:
	dh $@

override_dh_auto_configure:
	# Add here commands to configure the package.
	bash -c "cat ../SOURCES/*.patch | patch -p1"
	cmake -S . -B . -DCMAKE_INSTALL_PREFIX:PATH=/usr -DSYSCONF_INSTALL_DIR:PATH=/etc
	# --- end custom part for configure
```

The RPM spec file section for building and installing is just

```spec
%prep
%autosetup -p1 -n OpenBoard-1.7.0

%build
%cmake
%cmake_build

%install
%cmake_install

%fdupes -s %{buildroot}

%files
%license LICENSE
%doc COPYRIGHT
%{_datadir}/applications/%{fqname}.desktop
%{_datadir}/icons/hicolor/scalable
%{_datadir}/mime/packages/*
%{_datadir}/%{namelc}
%{_bindir}/%{namelc}
%config %{_sysconfdir}/%{namelc}
```

See also https://github.com/letsfindaway/OpenBoard/issues/99 for further discussions.


sebojolais:
I begin to test it. It works fine, thank you. Just one comment.
If we add this part at the end of CMakeList.txt:
`# Packages or Installer
include(InstallRequiredSystemLibraries)
set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
set(CPACK_PACKAGE_VERSION "${VERSION}")
set(CPACK_DEBIAN_PACKAGE_MAINTAINER "The OpenBoard team")
include(CPack)`

We could automatize the DEB and RPM generation by calling CPack in the command line: 
`cpack -G DEB`
or
`cpack -G RPM`
By this way the packaging should be easier.

letsfindaway:
Thank you for this proposal! Yes, of course! I started looking into `cpack`, as it would allow packaging for all platforms including Windows and MacOS as far as I have seen. IMHO this would be a really helpful feature.

I just tested your addition, and the generated RPM looked fine! The only difference to the packages I build on the Open Build Service are that those place the LICENSE and COPYRIGHT files to `/usr/share/licenses/OpenBoard` resp. `/usr/share/doc/packages/OpenBoard`, and that the Open Build Service uses `fdupes` to find identical files and to replace all but one by hardlinks. But this is just an optimization.

As I don't have any experiences with `cpack`, I have some questions:

- Which of those lines above are general, i.e. also for Windows and MacOS? In other words, what should be within some `if(LINUX)` etc., what can be outside?
- Which are specific to RPM or DEB?
- Where are variables like `CPACK_DEBIAN_PACKAGE_MAINTAINER` documented? I have not found that on https://cmake.org/cmake/help/latest/module/CPack.html.

sebojolais:
Unfortunately I have not lot of experiences with cpack. Here are my answers to your questions:

- These lines come from the [step 9 of the CMake tutorial](https://cmake.org/cmake/help/latest/guide/tutorial/Packaging%20an%20Installer.html). The tutorial is not platform specific. So these lines should work also for Windows, etc. (except CPACK_DEBIAN_PACKAGE_MAINTAINER) 
- Only CPACK_DEBIAN_PACKAGE_MAINTAINER is DEB specific. But it is ignored when the generator is not DEB. Because it is mandatory, I added it. But we have to use CPACK_PACKAGE_CONTACT instead of this DEB specific parameter in order to work with every generators.
- Have a look [here](https://cmake.org/cmake/help/latest/cpack_gen/deb.html?highlight=cpack_debian_package_maintainer#variable:CPACK_DEBIAN_PACKAGE_MAINTAINER)

letsfindaway:
Thanks for your pointers! I now added a "Packaging" section to the `CMakeLists.txt` file based on your proposal. I also mentioned you in the commit message. Hope this is ok?!

sebojolais:
Sure,it is fine.
I will test this PR on windows also and trying to support NSIS installer.

letsfindaway:
> I will test this PR on windows also and trying to support NSIS installer.

That would be fine! I don't have a Windows machine here, so I cannot test anything. But here are some notes on how I envision Windows (and MacOS) support.

- Add any necessary platform specific files to the section "Non-source file locations".
- There should not be the necessity to have a ThirdParty parallel repository. All such stuff should be added in the "Target and dependencies" section using `fetch_content`, so that these libraries are fetched and built together with OpenBoard as needed. I already started a little bit, but this section is not complete.
- The "Installation" section is also platform specific. Here everything is missing except Linux.

letsfindaway:
Superseded by #830

PR #705, opened 2022-11-10
workaround: avoid QTBUG-108284 with CSS styled cursors

letsfindaway:
See [QTBUG-108284](https://bugreports.qt.io/browse/QTBUG-108284)

- CSS styled cursors don't work with Qt 5.15 - 6.4 (at least)
- intercept CursorChange event on QWebEngineView's window
- forward to QGraphicsProxyWidget
- additionally mark other overriding functions as `override`
- commit may be reverted when bug is fixed in Qt

Fix was verified with Qt 5.15 and Qt 6.2 on Linux.

### DIFF
--- a/src/domain/UBGraphicsWidgetItem.cpp
+++ b/src/domain/UBGraphicsWidgetItem.cpp
@@ -42,7 +42,7 @@
 #include "api/UBWidgetUniboardAPI.h"
 #include "api/UBW3CWidgetAPI.h"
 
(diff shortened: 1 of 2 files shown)


kaamui:
Great finding and work, thank you !

letsfindaway:
> Great finding and work, thank you !

Thanks! And also a good example that discussions and mutual exchange of arguments are more productive than working alone. Thank you for your input and discussion.

PR #707, opened 2022-11-15
refactor: move non-config files out of etc

letsfindaway:
This small PR moves some of the files contained in `resources/etc` to a new directory `resources/template`. The reason behind this is a preparation for a FHS conformant package layout, where files from `resources/etc` will be installed to `/etc` or a subfolder of that.

Currently the `resources/etc` directory contains a mixture of configuration files and template files. Those should be separated.

The reason is, that package managers like `deb` and `rpm` handle files in `/etc` differently. While they assume, that most files installed by a package will never be changed by the user, they make an exception for files in `/etc`. These are configuration files and therefore expected to be modified. When installing a new version of a package, they can even leave modified files in their current state instead of overwriting them with the newly installed content, thus keeping any adaptation and customization applied by the user.

This small change prepares the OpenBoard directory layout for building such a FHS conformant package.

- It separates and moves template files from `resources/etc` to `resources/template`.
- It adapts the code accessing these resources.

Merging this PR will not affect the operation of OpenBoard in any way. It will however help the community to create packages better conforming to FHS and the rules of many distributions.


letsfindaway:
After a short walk with my dog :dog2: and reconsidering this PR, I now think that it is better to integrate this commit in to #698.

So I'm closing it here.

PR #712, opened 2022-12-02
background-grid-size-save

ThomasLucky13:
Saving background grid size when it has been changed.


habruening:
Was it not possible to use the variable crossSize directly?
Is the variable defaultCrossSize still needed?

PR #716, opened 2022-12-17
Handle paste or drop of web page URL to board

letsfindaway:
On the way to handling URLs pointing to web pages (content type is `text/html`) this PR also provides some fixes and refactoring as well as class documentation. It contains four commits:

### Fix: no drag operations after pasting a URL to an inaccessible resource

After pasting a URL to an inaccessible URL to the board, e.g. one which returns a 404 error, any operations involving dragging on the board have been blocked due to inappropriate handling of the error case. The first commit fixes this issue.

### Refactor: decouple embed parsing from web engine

The `QEmbedController`  and `QEmbedParser` had references to `QWebEngineView`. Architecturally this is not necessary. All these classes should do is to parse a piece of HTML and extract embeddable content, to control the embed dialog and to create the widgets.

This refactoring therefore moves these dependencies out of those classes. 

Additionally it adds some style changes and fixes some typos. Especially it does:

- Remove an unnecessary `#include "web/UBEmbedParser.h"` from `UBGraphicsScene`.
- Clean-up includes in `UBEmbedController.cpp`.
- Use the Qt5 style of `connect` using compile-time checked function pointers.
- Reorder function declarations and implementations, so that they have the same sequence in the header and the source.
- Add new public slots to set title and URL of web page.
- Add `const` to parameters and functions where possible.
- Always align `*` or `&` in function parameters to the type and not the name. Just a matter of taste, but we should be consistent.
- Always use braces after `if`, avoid single line bodies. This helps avoiding problems when later additional lines might be added.
- Avoid calling static function with an object.
- Use `#pragma once` instead of `#ifndef` guards. There was already a symbol used which did not match the class name.
- Include what you need instead of module includes like `QtGui`. See [Beware of Qt Module-wide Includes](https://www.kdab.com/beware-of-qt-module-wide-includes/) for reasons.
- Don't indent after first brace in class declaration.
- Abort parsing oEmbed when new HTML is provided.
- Avoid `clazy` warnings. They might indicate possible problems or performance drawbacks.

### Feat: Handle paste or drop of web page URL to board

After these preparations, implementation of this feature is easy:

- Add a mime type for `text/html` to `UBMimeType`.
- Add `UBEmbedController` to `UBBoardController` to provide the embed dialog and create the widget.
- Add `UBEmbedParser` to `UBBoardController` to parse the provided data.

### Docs: Documentation

`UBEmbedController` and `UBEmbedParser` have been annotated with Doxygen style documentation in the header. Just as a very first step...

### Further options

While implementing this I also thought about the following: 

- Currently the embed dialog does not have a "Cancel" button. It can however be cancelled using the "Close" button. But IMHO all dialogs should have a "Cancel" button if they can be cancelled.
- Currently a widget is always created in the library and as well placed on the board. Should there be options to only add it to the board or only add it to the library?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,4 @@ plugins/cffadaptor/objects
 ##########
 .directory
 .DS_Store
+Doxyfile
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -84,6 +84,9 @@
 
 #include "core/UBSettings.h"
 
+#include "web/UBEmbedController.h"
+#include "web/UBEmbedParser.h"
+
 #include "core/memcheck.h"
 
 UBBoardController::UBBoardController(UBMainWindow* mainWindow)
@@ -94,6 +97,7 @@ UBBoardController::UBBoardController(UBMainWindow* mainWindow)
     , mPaletteManager(0)
     , mSoftwareUpdateDialog(0)
     , mMessageWindow(0)
+    , mEmbedController(nullptr)
     , mControlView(0)
     , mDisplayView(0)
     , mControlContainer(0)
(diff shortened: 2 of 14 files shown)


kaamui:
@letsfindaway just saw that after copy pasting an url, the created object cannot be deleted ! In fact it's all its grey selection frame that cannot be interacted with ! 

As a workaround, I just go to the desktop mode and come back to board mode, then the object behaves as expected

letsfindaway:
> @letsfindaway just saw that after copy pasting an url, the created object cannot be deleted ! In fact it's all its grey selection frame that cannot be interacted with !
> 
> As a workaround, I just go to the desktop mode and come back to board mode, then the object behaves as expected

I can reproduce this. Can you open a new issue for that, here or on my project?

letsfindaway:
You can also interact with the object again if you e.g. click a link in the created web page or something like that. There seems to be some flag blocking events, which is not properly reset.

kaamui:
#774

Issue #640, opened 2022-06-23
Seyes background


sebojolais:
Hi,
In France, lot of teachers use [Seyes ruling](https://en.wikipedia.org/wiki/Ruled_paper#France) notebook.
For them, it makes sense to show the Seyes ruling by default as a background, without the necessity to add the 'Paper' application for each new page.

I did this modification in my side, and I think it is very useful. 
![Screenshot_20220623_120629](https://user-images.githubusercontent.com/5416255/175274636-4287b1bc-b5fb-4edf-95c6-ac2a068503b6.png)

If there are lot of Seyes ruling users that share this point of view, we can imagine this feature request :
- in order to keep OpenBoard simple, we can add just a seyes parameter in the configuration file to switch the ruled background in Seyes ruling. By this way, the GUI of OpenBoard is not changed. 
- If the Seyes parameter is TRUE, there are Plain background, Grid background, and Seyes ruled background. 
- If the Seyes parameter is FALSE, there are Plain background, Grid background, and ruled background.

If there is a consensus, I can try to provide a pull request
Best regards

sebojolais:
If someone needs this feature, it is implemented in this [branch](https://github.com/sebojolais/OpenBoard/tree/seyes_background) of the fork.
Due to no answer or comment by the community, I do not provide a pull request.

SeeDborg:
+1 on this request.

I am a trainer in France, and all the teachers regularly ask me for this feature. It would be really cool if it was added!

kaamui:
> If someone needs this feature, it is implemented in this [branch](https://github.com/sebojolais/OpenBoard/tree/seyes_background) of the fork. Due to no answer or comment by the community, I do not provide a pull request.

Hi,

sorry for the late reply. I find it interesting, and I could imagine merging it, as you provided a parameter to enable/disable it. 

Just a reminder of the conditions for me to merge pull request that don't fit a current need for Geneva schools : 

- the new feature can be enabled/disabled using the configuration file
- it is disabled by default
- if disabled, no changes in the current behavior of the app (and no regression)
- if enabled, it works :smile: 
- pull request provides minimal code modifications, (no refactoring unless mandatory, no debug comments or code left, no not related code, etc)

sebojolais:
Hi,
Thank you for keeping a place for community features that do not fit Genova schools requirements.
I follows every conditions and the pull request is here #727.
I will be very pleased to see this feature in future releases.
Thank you for your great job.

[...]

PR #727, opened 2023-01-06
Seyes background

sebojolais:
This pull request follows the #640 feature request.
The parameter introduced in the config file is located in 'Board' section, his name is 'SeyesRuledBackground' and it is a boolean.


kaamui:
Hi,

Even if it is not useful for Geneva schools, as seyes is not used at all here, your modification will be integrated to the official release. It will remain deactivated by default though.

PR #746, opened 2023-03-25
fix: consider image rotation

### LINKED ISSUE #744: Should respect picture orientation
Hi,

When importing JPEG files into the documents and into the board, orientation metadata is not respected. This is especially annoying as it is quite difficult inside OpenBoard to rotate a picture 90° or more (see #731) .

Regards,
Yvan

letsfindaway:
Currently transformation of images dropped on the board which is contained in image metadata (EXIF) is not considered by OpenBoard. This PR changes the way how images are loaded to consider such data.

- consider transformations contained in image metadata (EXIF)
- use QImageReader to control transformation

This PR fixes #744.


letsfindaway:
Obsoleted by #758 (merged).

PR #750, opened 2023-03-31
fix: PDF export when background PDF was scaled

letsfindaway:
I hope this PR now finally fixes the PDF export scaling. The main changes are

- use pageDpi to convert scene units to PDF units
- take the scaling between original PDF and scene into account
- use actual appearance on scene to calculate the scaling factor
- avoid rounding error in UBGraphicsScene::normalizedSceneRect

In fact this reintroduces a former idea, but in a different implementation. originally, the scaling of a PDF document on the scene was computed by getting `m11` of the transformation matrix. There were however cases - and I cannot exactly tell why - when this was not what we see on the screen. Now this factor is computed by using the scene bounding rect of the PDF and its original size as taken from the document.

When interpreting the scene coordinates, we now always use the pageDpi as stored in the document. In some former version (not the last one) the actual screen resolution was used, which of course can be different.

Additional a small rounding error was fixes when computing the normalized scene rect. The effect is barely visible, but for the sake of exactness..

I tested this now with all of the problematic documents we collected and it looks fine on my platform. Please test on your side before merging.

### DIFF
--- a/src/adaptors/UBExportFullPDF.cpp
+++ b/src/adaptors/UBExportFullPDF.cpp
@@ -216,6 +216,9 @@ bool UBExportFullPDF::persistsDocument(UBDocumentProxy* pDocumentProxy, const QS
 
             MergeDescription mergeInfo;
 
+            // factor between scene coordinates and PDF coordinates
+            double dpiScale = 72. / pDocumentProxy->pageDpi();
+
             int existingPageCount = pDocumentProxy->pageCount();
 
             for(int pageIndex = 0 ; pageIndex < existingPageCount; pageIndex++)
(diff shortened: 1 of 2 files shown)


letsfindaway:
@kaamui: What are the results of your tests of this PR?

letsfindaway:
> Thanks for the reminder.

You're welcome! I'm just checking what patches I still apply for my openSUSE builds. Now only four patches remain:

- one for the CMake build system I'm using
- one to suppress the software update check, which does not make sense when the package is updated by the distribution
- one for the configurable shortcuts, where I know that it will not make it into 1.7.0
- and one we have discussed before and were I still don't know what problems it may cause on MacOS (#569)

I'm quite happy with this status. What remains from my point of view are the translations, at least the update of the translation files using `lupdate`, so that contributors can base their work on these files.

kaamui:
On my side, 5 tickets remain, but probably just one more is going to reach the 1.7.0, and the others will be postponed. Translations come at the end of my process.

letsfindaway:
I extracted a list of changes between 1.6.4 and 1.7.0 from the commit messages. See https://github.com/letsfindaway/OpenBoard/wiki/Changes-in-OpenBoard-1.7.0. You may use this when compiling your list of changes for the release if you like.

kaamui:
It also comes at the end of my process ^^. 

It's really kind thank you ! I'll definitely use it !

Issue #214, opened 2019-03-27
Widget to embed live UVC image


hkiel:
I use ELMO document camera. Maybe it is possible to have a widget that can show the image of (any?) UVC camera. Best would be to also have a button to freeze the image and save it to the OpenBoard document for later reference.

cwilms:
I would also like to see UVC camera support. We use Hovercam document cameras and it would be helpful to annotate over the document camera feed. Have a way to freeze the live image and save it would be helpful too.

letsfindaway:
I think such a widget could be feasible for the upcoming version 1.7.0 of OpenBoard. Any volunteers? There are various examples on the Internet on how to access the camera with Javascript and play it on the screen, e.g. https://www.webdevdrops.com/en/how-to-access-device-cameras-with-javascript/. I assume such code could be wrapped in a widget.

Note that such a widget is not feasible for 1.6.4, as the browser engine used in this version does not support camera access.

letsfindaway:
A "Camera" widget is available in OpenBoard since version 1.7.0.

PR #757, opened 2023-04-12
Camera widget

letsfindaway:
This PR adds a Camera widget which allows to display a live camera image on the board as suggested by #214. Features include:

- display a live camera image on the board
- stop and continue by clicking on the image
- create snapshots in different sizes
- also usable as a fixed tool

To make this widget work, several fixes had to be applied to OpenBoard, which are contained here in separate commits:

- To allow using this as a tool, the `onenter` and `onleave` events also have to be created when the widget is a tool.
- The `sankore` widget API shall always work on the current scene, instead being fixed on the scene were the tool was created.
- There was still a problem with the position of drop down menus when a tool or a deselected widget was moved on the board.
- The handling of `data:` URLs was quite slow due to frequent conversions between `QString` and `QUrl`. The code also assumed, that a URL is short and e.g. stored those URLs as part of the page SVG. `data:` URLs can however have several megabytes. Some performance improvements have been made here.

The usage of the widget should be as simple and intuitive as possible:

- After starting the widget for the first time in an OpenBoard session, the user is asked to permit camera access.
- The widget is automatically resized to match the camera's aspect ratio.
- By clicking on the image, the user can freeze and unfreeze the live image. To visualize this feature, a transparent "play" or "pause" icon is displayed over the camera image, while the mouse is over the widget.
- When the mouse is over the widget, a snapshot button and a size selector is displayed in the top right corner. Here the user can select between S, M, L and XL snapshot size and then create an image from the current video.
- The widget can be converted to a tool, so that it is visible on each page. The snapshot is always placed on the current page.

### DIFF
--- a/resources/library/applications/Camera.wgt/config.xml
+++ b/resources/library/applications/Camera.wgt/config.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<widget xmlns="http://www.w3.org/ns/widgets" 
+        xmlns:ub="http://uniboard.mnemis.com/widgets"
+        id="http://letsfindaway.de/widgets/camera" 
+        version="0.1"
+        width="600"
+        height="400"
+        ub:resizable="true"
+        ub:transparent="false">
+    <name>Camera</name>
+    <author email="me@letsfindaway.de">Martin Winter</author>
+    <description>Live camera image with snapshot feature</description>
+    <content src="index.html"/>
+</widget>
--- a/resources/library/applications/Camera.wgt/css/style.css
+++ b/resources/library/applications/Camera.wgt/css/style.css
@@ -1,0 +1,104 @@
+body {
+  background-color: transparent;
+  overflow: hidden;
+  margin: 0;
+}
+
+html, body, .container {
+  margin: 0;
+  padding: 0;
+  height: 100%;
+  width: 100%;
+}
+
+img {
+  height: auto;
+  max-width: 100%;
+}
+
+video {
+  display: block;
+  width: 100%;
+  max-width: 100%;
+}
+
+.container {
+  display: table;
+}
+
+.cell {
+  display: table-cell;
+  margin: 0;
+  padding: 0;
+  background-color: #ccc;
+  text-align: center;
+  vertical-align: middle;
+}
+
+.content {
+  display: block;
+  width: 100%;
+}
+
+.button {
+  align-items: center;
+  background-color: #fff;
+  border-color: #dbdbdb;
+  border-width: 1px;
+  color: #363636;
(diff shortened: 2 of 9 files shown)


letsfindaway:
There will be an additional commit the next days including:

- Information when camera is not accessible
- Camera selection when more than one camera is connected
- Fix the missing application icon

letsfindaway:
The points above are now fixed.

letsfindaway:
The latest update to this PR now transfers the snapshot images as JPEG from the widget to the board. Together with #758 it also stores the images in JPEG format in the document, leading to much smaller documents than before, when PNG was used for all images.

letsfindaway:
If required or useful I can also split this PR into two:

- one for the fixes (event handling, drop-down position, URL conversions),
- another one for the widget.

Also other widgets like Papier will benefit from the fixes. Currently this widget is also affected by the problems fixed there:

- If you convert the Papier widget to a tool to apply the background to several pages, then it only operates on the original scene and does nothing after switching to another page (fixed with first commit).
- If you convert one of the interactivities to a tool, then the drop down menus e.g. for style are not correctly placed (fixed with first and second commit).
- After adding a background using Papier, the SVG file of that page contains an `<image>` with a `ub:source` attribute of 100kBytes or more, because the complete `data` URL is used as source. This is unnecessary overhead (fixed with third commit).

Please tell me whether splitting this PR would be useful.

kaamui:
The PR will be discussed with the team on 8th of May. If camera widget is not accepted immediately, I'll ask you to split the commits. 

One feedback maybe (I didn't look at it so sorry if I'm wrong) is that explicit size should be visible for user, instead of S, M, L and XL. What do you think ? Maybe something like `S (320px)`. OK it's not accurate as height is calculated according to the video size, but I think users are used to resolutions and could therefore have a better idea of what to expect. Or maybe update label at resize so you can display (320 x ??, 640 x ??, etc) ?

letsfindaway:
> The PR will be discussed with the team on 8th of May. If camera widget is not accepted immediately, I'll ask you to split the commits.

OK
 
> One feedback maybe (I didn't look at it so sorry if I'm wrong) is that explicit size should be visible for user, instead of S, M, L and XL. What do you think ? Maybe something like `S (320px)`. OK it's not accurate as height is calculated according to the video size, but I think users are used to resolutions and could therefore have a better idea of what to expect. Or maybe update label at resize so you can display (320 x ??, 640 x ??, etc) ?

That's just a tradeoff between simplicity (S, M, L) and transparency to the user and I'm happy to adapt this to whatever the team thinks is most usable. Also a dynamic update to the video size is feasible. My idea was just to avoid the complexity of a completely free user choice.

I'm waiting for the feedback from the team!

PR #758, opened 2023-04-15
perf: store images in source format

letsfindaway:
This PR addresses the storage consumption of OpenBoard documents containing images. 

Currently, all images are converted to PNG and stored as such. So JPEG images with a reasonable large resolution end up using much disk space.

This PR

- avoids conversion of any image to PNG, as it needs a lot of disk space,
- stores imported or dropped image files as their original data in the OpenBoard document,
- while OpenBoard versions before 1.7.0 can still read the documents.

The core change is a new function `addImage` in `UBGraphicsScene`, which shall replace `addPixmap` in as many cases as possible. Instead of a pixmap, it takes a `QByteArray` of the image data. All other parameters are the same. The data is stored to disk as is.

The code for importing an image, for reading and writing a document, for DnD handling and pasting is adapted accordingly.

This PR obsoletes #746, as it also contains this fix. Only one of these two PRs shall be merged, preferably this one.

### DIFF
--- a/src/adaptors/UBImportImage.cpp
+++ b/src/adaptors/UBImportImage.cpp
@@ -102,6 +102,7 @@ QList<UBGraphicsItem*> UBImportImage::import(const QUuid& uuid, const QString& f
     UBGraphicsPixmapItem* pixmapItem = new UBGraphicsPixmapItem();
     pixmapItem->setPixmap(pix);
     result << pixmapItem;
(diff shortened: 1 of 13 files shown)


letsfindaway:
Now also handling of images in the feature library is included. Additional small fixes in handling "move" in the library in second commit.

PR #764, opened 2023-05-15
Fix mixed problems

letsfindaway:
This PR fixes a series of problems listed in https://github.com/letsfindaway/OpenBoard/issues/130 related to some log messages:

- "Invalid path data" was caused by a very long SVG path in `Worldmap_wdb_splitted.svg`.
- "Uncaught TypeError" caused by the `index.html` of Hints and Tips.
- "Synchronous XMLHttpRequest", also caused by the `index.html` of Hints and Tips.
- "QIODevice::read: device not open" issued when dropping a widget to the board.
- "ERROR:command_buffer_proxy_impl.cc" when terminating the application.
- "Release of profile requested" when terminating the application.

For a deeper description of these errors see the issue linked above. Each commit in this PR fixes one or two of the problems.


src/core/UBPersistenceManager.cpp
```diff
@@ -144,10 +144,7 @@ void UBPersistenceManager::destroy()
 
 void UBPersistenceManager::onScenePersisted(UBGraphicsScene* scene)
 {
-    if (!mIsApplicationClosing) {
```
kaamui, line 147: why removing this check of `mIsApplicationClosing` ? Are you sure it's not going to end up with a double free corruption ?

letsfindaway: Yes, I'm quite sure that we don't have a double free. The sequence of calling this code is as follows:

- `onScenePersisted` is connected to `UBPersistenceWorker::scenePersisted` here: https://github.com/OpenBoard-org/OpenBoard/blob/4f49a2fa0e8feb626ca20957c9c9e2296dedf93a/src/core/UBPersistenceManager.cpp#L123
- This signal is emitted when the scene was persisted here: https://github.com/OpenBoard-org/OpenBoard/blob/4f49a2fa0e8feb626ca20957c9c9e2296dedf93a/src/core/UBPersistenceWorker.cpp#L78
- The scene object was passed to the `UBPersistenceWorker` in `UBPersistenceWorker::saveScene`, which is only called from this code: https://github.com/OpenBoard-org/OpenBoard/blob/4f49a2fa0e8feb626ca20957c9c9e2296dedf93a/src/core/UBPersistenceManager.cpp#L1170-L1173
- Here a copy of the scene is created for asynchronous saving. If we don't delete this copy in `onScenePersisted`, then no other code could do this, as no other code can have a pointer to the copied scene.

letsfindaway: I was also thinking about using a `shared_ptr` for scenes, because object ownership for scenes is not clearly defined. Mostly they are created and managed by the `UBSceneCache`, or by the `UBSvgSubsetAdaptor` on loading, where it is later inserted to the cache by the `UBPersistenceManager`. But there are also some exceptions, like the black scene for blackout or the transparent scene for desktop mode.

Scenes are passed around quite often, so when deleting a scene you can never be sure whether somebody still has a reference. Here a `shared_ptr` would clearly help.

A `shared_ptr` can be used, even if `QGraphicsScene` is a `QObject`, because `UBGraphicsScene` guarantees that there is no parent - except if somebody evil would call `setParent` afterwards.

I did not do this change for this PR, because I could clearly identify the two dangling scenes, and I wanted to keep the changes as small as possible. But if you think it is useful, I could also work on using `shared_ptr` for scenes.

BTW: Do you need somebody completing the German translation for 1.7.0, or is this done by your team?

kaamui: > I was also thinking about using a `shared_ptr` for scenes, because object ownership for scenes is not clearly defined. Mostly they are created and managed by the `UBSceneCache`, or by the `UBSvgSubsetAdaptor` on loading, where it is later inserted to the cache by the `UBPersistenceManager`. But there are also some exceptions, like the black scene for blackout or the transparent scene for desktop mode.
> 
> Scenes are passed around quite often, so when deleting a scene you can never be sure whether somebody still has a reference. Here a `shared_ptr` would clearly help.
> 
> A `shared_ptr` can be used, even if `QGraphicsScene` is a `QObject`, because `UBGraphicsScene` guarantees that there is no parent - except if somebody evil would call `setParent` afterwards.
> 
> I did not do this change for this PR, because I could clearly identify the two dangling scenes, and I wanted to keep the changes as small as possible. But if you think it is useful, I could also work on using `shared_ptr` for scenes.
> 
> BTW: Do you need somebody completing the German translation for 1.7.0, or is this done by your team?

It's funny because I already started implementing it yesterday, making the same change for UBGraphicsScene. It ended with a compile error in the svg subset adaptor (like if I was calling for a copy constructor of QGraphicsScene, didn't understand why) so I put it beside to go on on other subjects.

I'll share my modifications on a dedicated branch so you can join the effort.

kaamui: the branch is called `dev-shared-scene`

kaamui: > BTW: Do you need somebody completing the German translation for 1.7.0, or is this done by your team?

We actually have somebody, thank you.

kaamui:
Can you rebase it on actual `dev` state, so I can merge it ? Thank you !

letsfindaway:
> Can you rebase it on actual `dev` state, so I can merge it ? Thank you !

I removed the last commit also addressing the scene and rebased it to current `dev`.

PR #767, opened 2023-05-18
fix: shared_ptr in threads, circular references, others

letsfindaway:
Another PR to stabilize the usage of `shared_ptr`:

- break `shared_ptr` cycles in `UBGraphicsStroke`
- break `shared_ptr` cycles in `UBGraphicsCache`
- maintain ownership of `UBGraphicsCache` in `UBGraphicsScene`
- do not pass `shared_ptr` to other threads (`UBPersistenceWorker`)
- better handling of scene for `UBWidgetUniboardAPI`
- release scene after encoding in `UBPodcastController`

A detailed description is available here: https://github.com/letsfindaway/OpenBoard/issues/132#issuecomment-1553012031

### DIFF
--- a/src/domain/UBGraphicsWidgetItem.cpp
+++ b/src/domain/UBGraphicsWidgetItem.cpp
@@ -504,9 +504,8 @@ QString UBGraphicsWidgetItem::iconFilePath(const QUrl& pUrl)
     return file;
 }
 
-void UBGraphicsWidgetItem::activeSceneChanged()
+void UBGraphicsWidgetItem::initAPI()
 {
-    qDebug() << "Active scene changed, register/update API";
     registerAPI();
 }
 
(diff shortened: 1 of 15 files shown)


src/domain/UBGraphicsWidgetItem.cpp
```diff
@@ -510,6 +510,11 @@ void UBGraphicsWidgetItem::activeSceneChanged()
     registerAPI();
 }
 
+void UBGraphicsWidgetItem::setApiScene(std::shared_ptr<UBGraphicsScene>)
```
kaamui, line 513: scene seems to be unused, so if the purpose is just to call `registerAPI`, this `registerAPI` function should directly be called, isn't it ?

letsfindaway: Definitely, yes. I will change that.

letsfindaway: Removed unused `activeSceneChanged and renamed `setApiScene` to `initAPI` without parameters.

src/domain/UBGraphicsWidgetItem.cpp
```diff
@@ -504,9 +504,8 @@ QString UBGraphicsWidgetItem::iconFilePath(const QUrl& pUrl)
     return file;
 }
 
-void UBGraphicsWidgetItem::activeSceneChanged()
+void UBGraphicsWidgetItem::initAPI()
```
kaamui, line 507: Still, what would you create a public function called `initAPI` that calls `registerAPI`, rather than changing `registerAPI`'s visibility ? I don't see the interest.

kaamui: OK it makes sense.

PR #772, opened 2023-06-07
Fix compilation errors on win10-64x 

Messiaslogia:
I apologize for my last pull request because I put it on the "dev 1.7" branch instead of putting it on the "dev" branch.

All these executions were done following the wiki step by step. Using the "dev" branch as in the wiki. So I'm just going to duplicate the information from the last pull request I made on the wrong branch.

<--------------------------------------->

I ran into some errors while following the wiki steps to build OpenBoard 1.7-dev (https://github.com/OpenBoard-org/OpenBoard/wiki/Build-OpenBoard-on-Windows-10-(1.7))

The document specifies to use "Developer Command Prompt for VS 2019" in the Windows search bar" when using this prompt on my machine using the Win10 - 64x operating system returns me this error:

image

The error suggests that the libraries machine type 'x64' is in conflict with the target machine type 'x86'

I fixed this error using the "x86_x64 Cross Tools Command Prompt for VS 2019" prompt to do the entire process that is on the wiki and with it I was able to successfully compile openBoard and issue the nmake command in the project folder:

image

After the complete compilation of OpenBoard.pro I followed the wiki to execute the file "release.win7.vc9.bat" when following the execution it returned errors specifying that libraries were missing inside my Qt(5.15.2) installed:

image

I made a list of libraries that need to be downloaded for successful execution:

image

after that the installer was successfully compiled!

image

Updated the install page wiki on win - 1.7 - dev (it's in README.md)
Thanks for your service :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,175 @@
-# OpenBoard
-OpenBoard is an open source cross-platform interactive white board application designed primarily for use in schools. It was originally forked from Open-Sankoré, which was itself based on Uniboard.
+# Build OpenBoard on Windows 10 (1.7-dev)
 
-Supported platforms are Windows (10+), OS X (10.13+) and Linux (Ubuntu 20.04 is the only officially maintained platform, but many others are reported by the community to work).
+## ⚠️Disclaimer⚠️
 
-## Installing
+These instructions only work for the upcoming version : OpenBoard 1.7. At the time you are reading, you need to place yourself on the ```dev``` branch (explained below). If you want to build OpenBoard 1.6.4, you'll need to install QtWebKit instead of QtWebEngine, but we strongly recommend you to use the QtWebEngine library and the ``dev`` branch. If you still want to build OpenBoard 1.6.4, please refer to the instructions written for OpenBoard 1.6.4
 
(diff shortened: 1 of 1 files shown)


kaamui:
Thank you for your feedback. Indeed, the readme is misleading concerning the right terminal to use. I'll update it to "x86_x64 Cross Tools Command Prompt for VS 2019" as I'm pretty sure it's the one I use too. 

For ICU dlls, it's curious because it is supposed to be in the Qt lib folder. At least it is in Qt 5.15.0, Didn't try Qt 5.15.2 but I'll try to see what happens. Where did you download them then, in order to complete the release script ?

kaamui:
> Thanks for the feedback
> 
> I also tested the most current versions of Qt, but even then the ICU dlls were not present. I literally downloaded the dll files from the links I put in the README and manually placed them in Qt's lib folder.

Oh I think I know why : they indeed aren't there by default. I think I always had these dlls in the Qt bin folder because of previous use of QtWebkit. I'll try to see if they can be removed from iss or still needed. Thanks again for the feedback

PR #773, opened 2023-06-08
Fix broken Qt6 compatibility

letsfindaway:
Unfortunately I have broken Qt6 compatibility with two commits:

- 27ae0146f9407004e9b4c720e0ddc5b94d0c9bfa introduced the `leftRef` calls. These are not necessary and do no longer exist in Qt6.
- 7ba86ecdbece5b88770d614f2a3254c2b236d183 added an `enterEvent`, which has a different parameter type in Qt6. Due to the `override` annotation this was flagged as a compile error. I found some more occurrences of that problem which have not been flagged as a compile error because of the missing `override` annotation.

This PR fixes these issues, which only occur when compiling with Qt6.

- use `QString::left`, `mid` instead of `leftRef`, `midRef` as `QStringRef` is obsolete in Qt6
- use modified signature of `enterEvent`, parameter type is now `QEnterEvent*` instead of `QEvent*`

### DIFF
--- a/plugins/cffadaptor/src/UBCFFAdaptor.cpp
+++ b/plugins/cffadaptor/src/UBCFFAdaptor.cpp
@@ -36,8 +36,8 @@
 #include <QSvgRenderer>
 #include <QPainter>
 
-#include "UBGlobals.h"
 #include "UBCFFConstants.h"
+#include "core/UB.h"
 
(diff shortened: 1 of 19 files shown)


letsfindaway:
I will add a second commit extracting some version dependent types to a common header.

kaamui:
I was aware of this but also read that we should use QStringView instead.

Issue #794, opened 2023-08-30
Translations for "StartupHints" that are not in the `ts` files


kaamui:
Hi,

I just wanted to point out the community helping us with translations, that with the 1.7.0 comes new web pages welcoming the user in the app and providing some basic knowledge on "how to do basic things" in OpenBoard.

These pages need translation in every language possible, but contrary to the rest of the app, they are not referenced in `.ts`  files. The pages to translate are located [here](https://github.com/OpenBoard-org/OpenBoard/tree/master/resources/startupHints/locales)

A smarter way to handle translations in this directory should be possible, but I could not look at it so for now, every translation will need a full copy of the `en` folder (or any of the 4 officials folders `en`, `fr`, `de`, and `it`). 

Note that `it` and `de` are not up-to-date.

Please ask questions if you have any doubt on how to push a PR with these modifications. 

Thank you all for your work : you help OpenBoard (a free, open-source application for schools) to shine in every languages and countries we sadly cannot support more directly. 

There's not much we can do to reward your efforts except to thank you. Everyone helping with translations is constantly added to the "Credits" tab, in the preferences window of the application 🥇

Did I say thank you ? 😄 

Thank you !!

ovari:
@kaamui can you please create a `hu` folder and files (from en source) that translations can be entered? Thank you

kaamui:
Hi @ovari,

You can do it yourself ! Just go on the ``resources/StartupHints/locales` folder and copy-paste the `en` folder, then you just have to translate it and submit it via a PR !

ovari:
@kaamui we do everything in the Firefox web browser. Don't know how to copy and paste using the web browser. Can you please create `hu` folder? Thank you

kaamui:
@ovari you'll have to do it yourself, unfortunately. Please take example on  #798 for how to proceed.

letsfindaway:
If - for whatever reason - you do not want to clone the repository to your computer, you can still make everything in the browser:

- Create a fork.
- Switch to the `dev` branch by selecting it from the drop-down menu.
- Make sure your `dev` branch is up-to-date. If not, click "Sync fork":<br>
![grafik](https://github.com/OpenBoard-org/OpenBoard/assets/35591413/f24c324a-c309-4cd6-89fe-b6314a4bdd95)
- Click on the dropdown menu showing the `dev` branch. Enter a new branch name for your work and create that branch.
[...]

PR #805, opened 2023-09-28
Update OpenBoard_zh_CN.ts

aerowolf:


kaamui:
Hi,

Please note that there are other files to translate, that are not in the `.ts` file.

Please read #794 for more details.

Also note that 1.7.0 is about to be released (in approximatively two weeks).

PR #808, opened 2023-10-15
refactor: handle accented characters in web widgets

letsfindaway:
This PR proposes some improvements and refactorings for the handling of accented characters in web widgets

- make lookup maps `const static` instead of creating them for each instance of a web widget
- initialize `mLastDeadKey` in header file
  - avoids additional ifdef in the code
- handle two dead keys in a row
  - same key: write accent
  - different keys: drop both
- suppress blank after dead key
- use local variable for generated events to free memory afterwards
  - allocate an event if you send it via `QApplication::sendEvent`
  - but keep ownership if you pass it directly to an event handler
- fix key parameter when constructing `currentDeadKeyEvent`
- compact code by squashing consecutive `if` statements
- use `const` wherever possible
- use conversion `QString(const char*)` instead of `QString::fromUtf8(const char*)` as they have same behavior (see https://doc.qt.io/qt-5/qstring.html#QString-7)

Two points which might be done in addition:

- [ ] replace the reference to QTBUG-79216 by some other bug id, as this bug probably describes correct behavior (no `QKeyPressEvent` when input methods are active). I will try to provide a minimum proof-of-failure and create a QT bug report.
- [ ] move the static maps `sDeadKeys` and `sAccentedCharacters` and the function `getAccentedLetter()` out of the class and make them local static variables/functions in `UBGraphicsWidgetItem.cpp`. There is no need to expose this implementation detail in the header file, even not `private`. This is however not the current code style in OpenBoard.

### DIFF
--- a/src/domain/UBGraphicsWidgetItem.cpp
+++ b/src/domain/UBGraphicsWidgetItem.cpp
@@ -59,6 +59,81 @@
 bool UBGraphicsWidgetItem::sInlineJavaScriptLoaded = false;
 QStringList UBGraphicsWidgetItem::sInlineJavaScripts;
 
+#ifndef Q_OS_WIN
+/*
+ * workaround for a bug related to (at least) QTBUG-79216 - to be removed when bug is fixed
+*/
+const QMap<Qt::Key, QString> UBGraphicsWidgetItem::sDeadKeys
+{
+    {Qt::Key_Dead_Circumflex, "^"},   // ^ (Caret)
+    {Qt::Key_Dead_Grave, "`"},        // ` (Backtick)
+    {Qt::Key_Dead_Tilde, "~"},        // ~ (Tilde)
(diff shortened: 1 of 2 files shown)


src/domain/UBGraphicsWidgetItem.cpp
```diff
@@ -59,6 +59,81 @@
...
+    {"¨i", u8"ï"},
+    {"¨I", u8"Ï"},
+    {"¨o", u8"ö"},
+    {"¨O", u8"Ö"},
+    {"¨u", u8"ü"},
+    {"¨U", u8"Ü"},
+    {"¨y", u8"ÿ"},
+    {"¨Y", u8"Ÿ"}
+};
+#endif
```
kaamui, line 135: Good idea.

letsfindaway:
If it is useful I will provide the points listed above in a second PR, but as it is more about style and not about function this is not so urgent.

PR #809, opened 2023-10-16
UBCompassTool upgrade

ThomasLucky13:
Added new functionality to the compass.

изображение

1. Position and size normalization parameters. These parameters allow to align the position and size of the compass to the gridSize. Parameters are only enabled when the pageBackground is crossed.
изображение
изображение
изображение

2. Right angle rotation. This functionality allowsu to turn the compass counterclockwise to the nearest degree multiple 90 (0, 90, 180, 270).
изображение
изображение

3. Rotation of the compass by a given angle. You can rotate the compass by a given number of degrees absolutely or relative to the current rotation of the compass. At the same time, you can draw (this is optional).
изображение


kaamui:
Hi,

your work is cool, but it kinda takes the opposite direction of what our teachers are asking us. The main issue we want to fix for this tool is that user can't reduce the radius under a certain value. 

So I'll first try to find a way to rework the compass to make able to draw more little circles, before considering any other type of improvment.

PR #811, opened 2023-10-19
Fix glitches when moving items below live thumbnails

letsfindaway:
I think I finally found a solution for the problem with glitches when moving items below a live thumbnail, especially in large zoom factors. See https://github.com/letsfindaway/OpenBoard/issues/143#issuecomment-1771092503. We have to

- disable updates of board while updating thumbnail, and
- ignore the next repaint event on the board induced by re-enabling updates for thumbnail update.

### DIFF
--- a/src/gui/UBThumbnailWidget.h
+++ b/src/gui/UBThumbnailWidget.h
@@ -622,6 +622,7 @@ class UBDraggableLivePixmapItem : public UBDraggableThumbnailItem
         QTransform mTransform;
         QTimer updateTimer;
         int updateCount;
+        bool ignoreNextEvent;
 };
 
 namespace UBThumbnailUI
--- a/src/gui/UBThumbnailWidget.cpp
+++ b/src/gui/UBThumbnailWidget.cpp
@@ -37,6 +37,7 @@
 #include <QWidget>
 
 #include "board/UBBoardController.h"
+#include "board/UBBoardView.h"
 
 #include "core/UBSettings.h"
 #include "core/UBApplication.h"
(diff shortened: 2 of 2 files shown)


src/gui/UBThumbnailWidget.h
```diff
@@ -622,6 +622,7 @@ class UBDraggableLivePixmapItem : public UBDraggableThumbnailItem
         QTransform mTransform;
         QTimer updateTimer;
         int updateCount;
+        bool ignoreNextEvent{false};
```
kaamui, line 625: @letsfindaway Shouldn't we maintain a consistency on how data members are initialized ?

letsfindaway: My idea was: if I add some code, then I already try to stick to what we want for a refactored OpenBoard. And Initializing members in the header instead of each constructor is bad according to https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers. I won't change existing code in a bug fix, but try to use a better style for all things I add.

But to keep consistency I can also move this to the constructor. It's just a trade-off.

letsfindaway: Just tell me what you want to have here.

kaamui: I understand. I personally think it should remain consistent until we refactor things, but each pov is arguable. 

My main concern is more that I'm not sure curly braces should be preferred to `=` for a single argument initialization. Ok it is what is shown in the C++ core guidelines, but I find it un-intuitive, and to my knowledge it doesn't makes any difference in this case. If there are no good reason to prefer `{42}` to `=42`, please use the latter.

see : https://herbsutter.com/2013/05/09/gotw-1-solution

kaamui:
I was probably not clear enough, but I was OK with your first initialization, but with `=`, except if you had good reasons to use curly braces for this primitive's initialization. 

To not make you waste more time, I'll merge it as-is.

PR #814, opened 2023-10-24
fix: sporadic crash due to dangling connection

letsfindaway:
During my tests for the issue with the live thumbnails I got some sporadic crashes while zooming. I now could catch two occurrences and I think I can trace them back to dangling connections pointing to deleted objects. See https://github.com/letsfindaway/OpenBoard/issues/147.

Here is my fix:

- connections to a lambda are only automatically deleted if
  - the sender is deleted or
  - a context parameter is given in the connect call before the lambda
- use `clazy` to find all missing context parameters
- add missing context parameter to lambda connections
- note: not all fixed lines are "dangerous"
  - as many fulfill the first condition
  - they are deleted when the sender is deleted


letsfindaway:
Instructions on how to reproduce the problem: https://github.com/letsfindaway/OpenBoard/issues/147#issuecomment-1778590878

PR #824, opened 2023-11-13
feat: Use pkg-config on Linux

Vekhir:
This is mainly motivated by the Arch Linux package of `openboard-git` which has to carry a [patch](https://aur.archlinux.org/cgit/aur.git/tree/quazip.patch?h=openboard-git) to define where `quazip` is found, since Arch Linux packages `quazip` differently from Ubuntu. The use of `pkg-config` makes this patch redundant.

While at it, I also cleaned up the linking for `podcast` and `pdfMerger`, utilizing `pkg-config` and removing unnecessary steps. Especially the commit message for `podcast` goes into great detail as to the reasoning.

Together, there shouldn't be any change in behaviour (except requiring `pkg-config` be installed). In turn, it greatly improves the infrastructure and removes some burden from packagers downstream.
It would be great if this set of patches could make it into 1.7.

This work is partially based on #651.

-- Vekhir

### DIFF
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -464,17 +464,36 @@ linux-g++* {
     LIBS += -lcrypto
     #LIBS += -lprofiler
     LIBS += -lX11
+    CONFIG += link_pkgconfig
+    PKGCONFIG += poppler
 
+    # search and configure quazip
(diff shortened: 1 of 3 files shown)


Vekhir:
Essentially done in #830 which got merged today.

PR #825, opened 2023-11-13
feat: Automatically create translations using lrelease

Vekhir:
Since [QT 5.12](https://code.qt.io/cgit/qt/qtbase.git/tree/dist/changes-5.12.0/?h=v5.12.0), `CONFIG` supports [`lrelease`](https://doc.qt.io/qt-6/qmake-variable-reference.html#config) to include the compiling of translations in the `make` process. This means that
1. `lrelease` doesn't need to be run separately
2. The command is independent of the location of the `lrelease` executable. This is important as Qt6 on Arch Linux puts the executable in a custom directory under `qt6/bin` where it can be found by Qt, but running separately requires changing `$PATH`.

By default, the results are placed in `release/`, and while there is the variable [`QM_FILES_INSTALL_PATH`](https://doc.qt.io/qt-6/qmake-variable-reference.html#qm-files-install-path), it doesn't seem to change that path.

Do you have experience with the config option `lrelease` and `QM_FILES_INSTALL_PATH` to set the result path to `resources/i18n`? I haven't been able to.

-- Vekhir

### DIFF
--- a/OpenBoard.pro
+++ b/OpenBoard.pro
@@ -3,6 +3,7 @@ TEMPLATE = app
 
 CONFIG += c++17
 CONFIG -= flat
+CONFIG += lrelease
 CONFIG += debug_and_release \
           no_include_pwd
 


letsfindaway:
Hi,

I have seen your PRs, aiming to built the package for Arch Linux. As you have already seen, I was first working on a restructuring of the `qmake` based builds, but now also proposed a `cmake` based build system.

Just out of curiosity I tried to create a PKGBUILD file using this build system and added this now to my experimental builds here: https://build.opensuse.org/package/show/home:letsfindaway:experimental/OpenBoard.

The Arch build is currently blocked due to three packages which were not available on the chosen mirror, but I expect that this will resolve soon. On my local machine I already had a successful build.

Just for testing, I added just the minimum patches for this built. You may also test to add the others.

Vekhir:
> To download the source as .tar.gz just copy the "Download ZIP" link and replace .zip by .tar.gz. Then GitHub is delivering also > this format. I had to tweak the download however, as we currently don't have a 1.7.0 tag.
>
I meant your version on build.opensuse.org. Anyway, making a git version is a standard process and as such rather quick to do.

> Yeah. On my local machine I have installed both, Qt5 and Qt6. Here I need a "manual" selection of the Qt version. So I wanted to tell explicitly what Qt version to use.
>
> Edit: but perhaps something like: if Qt5 not found then try Qt6 would be helpful.

I found this: https://doc-snapshots.qt.io/qt6-dev/cmake-qt5-and-qt6-compatibility.html#supporting-older-qt-5-versions. Perhaps for your use case we could first check whether `QT_VERSION_MAJOR` has been set like this:
```
if(NOT DEFINED QT_VERSION_MAJOR)
    find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Core)
endif()
```
`QT_VERSION` would be renamed to `QT_VERSION_MAJOR`. What do you think?

At least for Arch packages, it is customary to always build in a `chroot`, i.e. an isolated environment, so mixing Qt5 and Qt6 shouldn't happen.

Vekhir:
> Your pkgver is 1.6.4, but you reference webengine? This version was still using webkit.

For `-git` packages, the pkgver works a bit differently - the 1.6.4 derives from the fact that there is no 1.7.0 tag. `makepkg` will always fetch the latest commit from `master`, so the version is effectively 1.7.0rc-231102.
The full pkgver `1.6.4.r502.g1420d5f3` can be read as `Based on commit 1420d5f3 which is 502 commits ahead of tag 1.6.4`

letsfindaway:
> > Your pkgver is 1.6.4, but you reference webengine? This version was still using webkit.
> 
> For `-git` packages, the pkgver works a bit differently - the 1.6.4 derives from the fact that there is no 1.7.0 tag. `makepkg` will always fetch the latest commit from `master`, so the version is effectively 1.7.0rc-231102. The full pkgver `1.6.4.r502.g1420d5f3` can be read as `Based on commit 1420d5f3 which is 502 commits ahead of tag 1.6.4`

Thanks for the clarification!

Our discussion is now far away from the title of this PR. I have now created a discussion on my fork at https://github.com/letsfindaway/OpenBoard/discussions/155 which could probably be a place to collect more ideas and improve the build for various distributions. Feel free to contribute!

Issue #828, opened 2023-11-22
Button positioning and resizing in various widgets


Vekhir:
When zooming in and out, the size and position of the close/rotate/resize buttons changes inconsistently between widgets, in some cases even inconsistently between the actual actionable area and the SVG image.
This issue is a list of instances where such inconsistencies were found. If they are resolved or considered intentional, I will mark them as done. I haven't found a previous list.
Making the buttons relative to their widget is in my opinion a good starting point. Deviations can be discussed in the comments.

Ruler:
- Close button:
    - The close button image size is absolute
    - The close button action area is absolute
- Rotate button:
    - The rotate button image size height is absolute if zoom level is < 1.0
    - The rotate button action area is absolute
    - For larger zoom levels, the action area and the image disconnect entirely (i.e. the area is outside the image).

Protractor:
All buttons scale consistently relative to their widget.

Compass:
- Close button:
    -  The close button image size is absolute
    -  The close button action width is absolute
    -  The close button action height is relative
- Resize button:
    - The resize button image size is absolute
    - The resize button action width is absolute
    - The resize button action height is relative

Triangle:
- All button images scale absolutely.
- All button action areas scale relatively to their widget.
- Mirror buttons:
    - If the height of the widget is too small (measured with the ruler at <5 cm for the upper and <4cm for the lower mirror button), then the image disappears, but the action area still exists and the triangle can be mirrored.
 - Reordering the buttons if the triangle gets small might be an improvement.

Drawings:
- The buttons and the border are absolute, but I think that's mostly fine.
- Rotate button:
    - The button image moves into the frame and out of the grey border for increasing zoom levels.
    - The pointer change is always above the image, suggesting the action area is the same, but oftentimes, the drawing is moved instead of rotated at increasing zoom levels.

Leave your thoughts below. For extended discussion on a subset of changes it is advisable to create another issue for that topic and link it here.

letsfindaway:
Thanks for the detailed analysis.

See also https://github.com/letsfindaway/OpenBoard/issues/146.

I think we need a general decision how buttons should behave:

- Should they zoom together with the item?
- What if they don't fit for very small zoom levels?

PR #829, opened 2023-11-23
Improve Ruler widget and buttons

Vekhir:
The ruler background was a normal rectangle compared to the rounded border. The background is now rounded too and the outline that is used for the border is also used for the background, so they are now directly linked.

Fixed the points under `Ruler` in #828. The ruler button layout now looks consistent with non-overlapping elements even at high or low zoom levels.


Vekhir:
@letsfindaway @kaamui 
This patch to refine the Ruler widget is now rebased against the current `dev` branch. Can you take a look at it?

kaamui:
Hi @Vekhir,

Thank you for your help and contribution on OpenBoard. 

Yes it's planned ^^

Vekhir:
Great!

letsfindaway:
I just want to add that more work has to be done similar to that. The resize behavior of tool buttons is quite inconsistent. See https://github.com/letsfindaway/OpenBoard/issues/146.

PR #830, opened 2023-11-24
Cmake community builds

letsfindaway:
This PR is a new attempt to ease Linux community builds. It supersedes

- #651 
- #698

This PR only adds minimal changes to the existing code and build system, which do not affect building OpenBoard as `qmake` project for Windows, MacOS and Linux. It then provides an **additional** `cmake` based build system, which does not intend to replace the existing `qmake` project, but instead provides an alternative way to easily build FHS-compliant packages for various Linux distributions. The target of this work is

- Focus on Linux in the first place.
- Work on as many distributions as possible with minimal customization.
- Provide an FHS-compliant file system layout.

The `cmake` builds are known to work with 

- openSUSE Leap 15.4, 15.5, 15.6 and Tumbleweed
- Ubuntu 22.04, 22.10, 23.04 and 23.10
- Debian 11, 12, 13 (Testing) and 14 (Unstable)
- Arch

but will most probably also work for many other distributions.

This PR consists of three commits:

- First, some code adaptations provide a way to configure application and configuration directories as needed for FHS-compliant builds. These changes will not affect `qmake` based builds.
- The second commit adds the `cmake` files and some supporting files. It does not change the existing code base.
- The third commit extracts the version information from the `qmake` file to a new file `version.txt`, which can be used both by `qmake` and `cmake`. It guarantees version consistence between those two.

For a discussion about this work see https://github.com/letsfindaway/OpenBoard/discussions/155.

This PR not only helps community maintainers and users, who want to build OpenBoard on an arbitrary Linux distribution. It also provides an early indicator for possible incompatibilities with several, also brand new library versions. In contrast to the previous, now superseded PRs, it does not change the project builds for Geneva schools in any way.


letsfindaway:
Thanks for merging. I have now set up a wiki page describing the use at https://github.com/letsfindaway/OpenBoard/wiki/CMake-builds.

You may either copy this page or link to it (then I can still edit it). This page references another page listing package names for dependencies: https://github.com/letsfindaway/OpenBoard/wiki/Package-dependencies. As I'm still editing this page, I suggest just to link it and not to copy it.

PR #869, opened 2024-01-13
fix: fix incorrect sRGB profile

letsfindaway:
This PR avoids some warnings in the log file about known incorrect sRGB profiles. It has no influence to anything the user will see.

- determine files by running `find . -name "*.png" -exec pngcrush -n -warn {} \;`
- fix files by running `mogrify` over each incorrect file
- some files have to be fixed with `pngcrush` to avoid warning about "New bKGD index = 0"
See also #539 and https://github.com/letsfindaway/OpenBoard/issues/154


soudjoud:
Tuutluuò-t	e	fjòyuàgdgeu ry

PR #888, opened 2024-02-11
Flexible ruler description

letsfindaway:
This PR replaces the fixed background rulings (blank, lined, crossed, Seyes) by a mechanism that allows flexible ruler descriptions, where additional background rulings can simply be added by providing an additional description file. In the `resources/etc/background` directory, there is already a set of 12 different rulings (plus one for the blank page).

The idea is to have an XML description for the rulings. There is an XML Schema file, which describes the format both formally and with user readable comments. Basically a background ruling is described as a set of line groups. Each line group has a defined angle and one or several lines in a specified distance. Line groups may repeat infinitely. The color and opacity of lines can be specified, but also the default colors defined in the preferences can be used. Additional settings allow to limit lines to the right and the left and to add a border on one or both sides.

For the user, the background palette changes as follows:

- It now displays a set of 6 different background rulings.
- The selection of light or dark background color is now a switch.
- In the preferences, the user may select which of the available rulings are presented in the palette and in which order by simply dragging them to the desired position.

Background rulings are stored in the document SVG files. To provide some interoperability with existing documents, the current flags for "lined", "crossed" and "intermediateLines" are interpreted, when no ruling description is provided. These flags are also written when rulings are used, which to some extend mimic the "old" patterns.

This PR is still a draft, as I want to collect more input about usability. The code may also need more testing.


letsfindaway:
## TODOs

Here's a list of TODOs for this PR:

- [ ] Move background button templates to resources.
- [ ] Store unknown background pattern locally.
- [ ] Allow to delete local background patterns by dragging them to a trash icon displayed at the end of the list in the preferences dialog.
- [ ] Write a Wiki page with user instructions and a guide to create own background patterns.

sebojolais:
Just a comment:
The Seyes background feature introduced in 1.7.0 was more a trick than a feature.
With this PR, it will be a real feature. So I think this PR is welcome. 
+1

letsfindaway:
Now I think the PR is good enough to remove draft status. Thanks @sebojolais for valuable input!

sebojolais:
Another good point of this PR: it is not a local feature, it will benefit for every users.

ch-westphal:
Very useful feature! Another interesting option would be to have a specific logo in the corner of the page.

kaamui:
Hi @letsfindaway,

do you anticipate any issue between this feature and the snap-to-grid one ?

letsfindaway:
This PR is now updated to deal with the snap-to-grid feature. We can now snap to a flexible background definition as follows:

- If a background only contains lines, then we snap to the closest line.
- If a background has any grid pattern, then we snap to all intersection points of the grid.
- Border definitions are not used for snapping.
- If a background has a margin, then we do not snap when objects are on that margin.
- The blank background is defined to have an invisible 10mm grid, so that it behaves as before.

kaamui:
I remember some features requests that asked for the ability to have a dark green background to mimic some old school blackboards where the slate used was of this color. Do you think it would be possible to add this as a parameter, where one must give the background color, and the expected behavior (behave as a dark background, or behave as a light background) ?

letsfindaway:
> I remember some features requests that asked for the ability to have a dark green background to mimic some old school blackboards where the slate used was of this color. Do you think it would be possible to add this as a parameter, where one must give the background color, and the expected behavior (behave as a dark background, or behave as a light background) ?

The background color is still drawn here:
https://github.com/OpenBoard-org/OpenBoard/blob/4dd51f4c39a572ae5690fda0cd1578f90bb08eae/src/domain/UBGraphicsScene.cpp#L2913-L2920

So the flexible ruler description does not affect that in any way. As long as you replace black and white by configurable colors there, it will work. And of course, the dark background should really be dark and the light background should be light, so that the foreground color substitution works.

This means, that the feature of a configurable background color is completely independent of the rulings!

letsfindaway:
> Good point, it's not mandatory to attach this to the ruler descriptions.

Yes, exactly. The ruler descriptions currently just assume that there are exactly two possible background colors, light and dark. And for those they offer the option to define a color. But also this is only an option. If you omit that, then the line color defined in the preferences is used.

The color definition in the ruling description is more meant for things like the red lines in Seyes or similar.

PR #892, opened 2024-02-13
Improve zLevel handling

kaamui:
As noticed by @letsfindaway, the way available ZLevel values are searched is extermely bad in term of performances. This pull request is a first draft to pursue analysis, improvements and refactoring of this handling

### DIFF
--- a/src/domain/UBGraphicsScene.cpp
+++ b/src/domain/UBGraphicsScene.cpp
@@ -138,6 +138,8 @@ qreal UBZLayerController::generateZLevel(QGraphicsItem *item)
         result =  generateZLevel(type);
     }
 
+    mAlreadyUsedZLevels.insert(result);
+
     return result;
 }
 
@@ -291,6 +293,8 @@ void UBZLayerController::shiftStoredZValue(QGraphicsItem *item, qreal zValue)
 {
     itemLayerType::Enum type = typeForData(item);
 
+    mAlreadyUsedZLevels.insert(zValue);
+
     if (validLayerType(type)) {
         ItemLayerTypeData typeData = scopeMap.value(type);
         if (typeData.curValue < zValue) {
@@ -302,18 +306,9 @@ void UBZLayerController::shiftStoredZValue(QGraphicsItem *item, qreal zValue)
 /**
  * @brief Returns true if the zLevel is not used by any item on the scene, or false if so.
  */
-bool UBZLayerController::zLevelAvailable(QGraphicsItem* item)
+bool UBZLayerController::zLevelAvailable(const qreal zLevel) const
 {
-    foreach(QGraphicsItem* it, mScene->items())
-    {
-        if (item != it)
-        {
-            if (it->zValue() == item->zValue())
-                return false;
-        }
-    }
-
-    return true;
+    return mAlreadyUsedZLevels.contains(zLevel);
 }
 
 UBGraphicsScene::UBGraphicsScene(std::shared_ptr<UBDocumentProxy> document, bool enableUndoRedoStack)
@@ -1954,14 +1949,15 @@ void UBGraphicsScene::addItem(QGraphicsItem* item)
     // the default z value is already set. This is the case when a svg file is read
     if(item->zValue() == DEFAULT_Z_VALUE
             || item->zValue() == UBZLayerController::errorNum()
-            || !mZLayerController->zLevelAvailable(item))
+            || !mZLayerController->zLevelAvailable(item->zValue()))
     {
         qreal zvalue = mZLayerController->generateZLevel(item);
         UBGraphicsItem::assignZValue(item, zvalue);
     }
-
     else
+    {
         notifyZChanged(item, item->zValue());
+    }
 
     if (!mTools.contains(item))
       ++mItemCount;
--- a/src/domain/UBGraphicsScene.cpp
+++ b/src/domain/UBGraphicsScene.cpp
@@ -308,7 +308,7 @@ void UBZLayerController::shiftStoredZValue(QGraphicsItem *item, qreal zValue)
  */
 bool UBZLayerController::zLevelAvailable(const qreal zLevel) const
 {
-    return mAlreadyUsedZLevels.contains(zLevel);
+    return !mAlreadyUsedZLevels.contains(zLevel);
 }
 
 UBGraphicsScene::UBGraphicsScene(std::shared_ptr<UBDocumentProxy> document, bool enableUndoRedoStack)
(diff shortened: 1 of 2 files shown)


src/domain/UBGraphicsScene.cpp
```diff
@@ -302,18 +306,9 @@ void UBZLayerController::shiftStoredZValue(QGraphicsItem *item, qreal zValue)
...
-    {
-        if (item != it)
-        {
-            if (it->zValue() == item->zValue())
-                return false;
-        }
-    }
-
-    return true;
+    return mAlreadyUsedZLevels.contains(zLevel);
```
letsfindaway, line 311: This should 

`return !mAlreadyUsedZLevels.contains(zLevel);`

The level is available if it is **not** contained in the set.

src/domain/UBGraphicsScene.cpp
```diff
@@ -138,6 +138,8 @@ qreal UBZLayerController::generateZLevel(QGraphicsItem *item)
         result =  generateZLevel(type);
     }
 
+    mAlreadyUsedZLevels.insert(result);
```
letsfindaway, line 141: Do we also have to delete a z level from the set when an item is deleted?

letsfindaway:
I also do not understand what `changeZLevelTo` actually does and how this would affect our set of z levels. Finally I think there are more places where the z level of an item is changed and therefore the set needs to be changed. So instead of modifying the map when a z level is generated I propose to do this when an item's z value is set or modified, i.e. whenever `UBGraphicsItemData::ItemOwnZValue` is modified and/or `item->setZValue()` is called. This is in

- `UBBoardController:722`
- `UBItem:74` (the occurrence in the board controller should be replaced by a call to this function)

However this function operates on the item, where the `UBZLayerController` is a private member of the scene. I don't have a good idea how this can be done in a well structured way.

src/domain/UBGraphicsScene.cpp
kaamui: Should we reassign everything every time a zValue is changed (e-g item with zLevel 99567 is removed, should item with zLevel 99566 chenge to zLevel 99567) ?

Or we don't do it, until item is grouped or user actively changes zLevel ?

kaamui:
> I also do not understand what `changeZLevelTo` actually does and how this would affect our set of z levels. Finally I think there are more places where the z level of an item is changed and therefore the set needs to be changed. So instead of modifying the map when a z level is generated I propose to do this when an item's z value is set or modified, i.e. whenever `UBGraphicsItemData::ItemOwnZValue` is modified and/or `item->setZValue()` is called. This is in
> 
>     * `UBBoardController:722`
> 
>     * `UBItem:74` (the occurrence in the board controller should be replaced by a call to this function)
> 
> 
> However this function operates on the item, where the `UBZLayerController` is a private member of the scene. I don't have a good idea how this can be done in a well structured way.

I like the idea that the scene handles it, via its ZLayerController.

kaamui:
There is also this static function : `void UBGraphicsItem::assignZValue(QGraphicsItem *item, qreal value)` 

But I don't think it's the good way to do it. A `QGraphicsItem` has no idea (and doesn't have to) what should be its zValue along the user actions on the scene. I don't know if it works for any practical use cases, but the scene should be responsible to the handling of item's layers. 

Another issue I see (maybe it's not an issue) is that `copyItemParameters` assigns the same zValue to the copy.

src/domain/UBGraphicsScene.cpp
kaamui: > Do we also have to delete a z level from the set when an item is deleted?

If I remove an object that was one of the first I put on the scene (so it's behind other objects), I don't want this new Item I'm adding to be put behind older objects. I we remove zLevel form the QSet, we should reasign other values, so there's no available "hole" in the set.

letsfindaway:
> There is also this static function : `void UBGraphicsItem::assignZValue(QGraphicsItem *item, qreal value)`
> 
> But I don't think it's the good way to do it. A `QGraphicsItem` has no idea (and doesn't have to) what should be its zValue along the user actions on the scene. I don't know if it works for any practical use cases, but the scene should be responsible to the handling of item's layers.

Of course the `QGraphicsItem` **has** an idea about its z value. It is a property of it. It is indeed the only authoritative source of this value. The scene just uses it.

> 
> Another issue I see (maybe it's not an issue) is that `copyItemParameters` assigns the same zValue to the copy.
> 

This is ok. Copying is used when saving a scene. And the copy should have the exact same value.

> So another question is "should we avoid items to have the same zValue ?". In terms of user needs, I think it's cool to have this zLevel handling, but maybe for 99% of the use cases, we just want to annotate our PDF and we don't care, as long as the PDF is behind the annotations.

We also have to manage the z value e.g. of overlapping images etc. So it is essential.

> 
> So if we can avoid reassignments each time an action is performed, (even if it's probably no big deal for a QSet of qreal), maybe it's better.

But another perspective on loading the scene, which is our initial problem:

In SVG, there is no z value. Items are painted in the sequence they occur in the scene. This is guaranteed when saving a scene by
https://github.com/OpenBoard-org/OpenBoard/blob/a3233be65ffbad0d680c614ded5e9e2e9eee2d16/src/adaptors/UBSvgSubsetAdaptor.cpp#L1226-L1228

So when we load a scene we can safely assume that the z values are incremental. So for me the question is 

- whether we have to save the z value at all in an attribute
- and whether we have to check the value when loading.

So another idea: leave the performance-wise bad implementation as it is, but do not call this function when loading. Just assume it is all right. And if two items would have the same value: would it hurt? Qt would then also paint them in the order of insertion.

kaamui:
> This is ok. Copying is used when saving a scene. And the copy should have the exact same value.

Not only. When an item is duplicated, the zValue is copied too. As I also wrote later, maybe it's not an issue, I wanted to point out this part of the puzzle.

kaamui:
> > So another question is "should we avoid items to have the same zValue ?". In terms of user needs, I think it's cool to have this zLevel handling, but maybe for 99% of the use cases, we just want to annotate our PDF and we don't care, as long as the PDF is behind the annotations.
> 
> We also have to manage the z value e.g. of overlapping images etc. So it is essential.

So we agree that we should not avoid two items to have the same zValue.

letsfindaway:
> > > So another question is "should we avoid items to have the same zValue ?". In terms of user needs, I think it's cool to have this zLevel handling, but maybe for 99% of the use cases, we just want to annotate our PDF and we don't care, as long as the PDF is behind the annotations.
> > 
> > 
> > We also have to manage the z value e.g. of overlapping images etc. So it is essential.
> 
> So we agree that we should not avoid two items to have the same zValue.

At least not while loading. We probably need that when we move an item up or down, but then it is not a performance issue.

PR #894, opened 2024-02-17
The grid size is now stored in the configuration.

habruening:
At startup of the application, the grid is calculated differently on different systems. The intended behaviour is that 1cm on screen is one grid space. That works perfectly. But in practice this is not often needed. And unfortunately the behaviour has a negative side effect: If you open the same file on different systems with different screen configurations (which is a typical use case for teachers), and create new pages, these pages get a another grid size than the already existing pages. That is a bit confusing. And especially when the ruler is used the behaviour is not nice, because on different systems it measures different sizes.

I propose to change the behaviour as follows.

The grid size is remembered. The grid size changes only when the user explicitly sets a new value with the slider. At next startup this value is kept. The nice 1cm calculation only happens when the user clicks on the "set default size" button. The start value for the grid is 40, because that value should be comfortable for most people.

I implemented this. The code simplifies a bit, because now we don't have to store the default value any more. Interestingly the connection to the signal UBDisplayManager::screenRolesAssigned becomes obsolete. It makes no sense any more to automatically update the grid size, when the screen configuration changes.

### DIFF
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -153,10 +153,6 @@ void UBBoardController::init()
 
     setActiveDocumentScene(doc);
 
-    connect(UBApplication::displayManager, &UBDisplayManager::screenRolesAssigned, this, [this](){
-        initBackgroundGridSize();
-    });
-
     undoRedoStateChange(true);
 
 }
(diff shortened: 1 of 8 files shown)


ThomasLucky13:
Hello! A similar solution to the problem has already been proposed earlier in this PR: https://github.com/OpenBoard-org/OpenBoard/pull/712

PR #895, opened 2024-02-18
Remove Masking in Desktop mode

habruening:
I fixed the problem that you cannot click through the board in desktop mode. For me it works now. But I can only test it on X11 Linux. All this is very platform dependent. Someone must test this on Windows, Wayland Linux and MacOS, with Qt6, in the presenter mode and carefully review my changes. I think, my commit is quite a risky thing. So I don't know if and how you developers prefer to change such things. At some time this thing must be fixed.

Honestly I would recommend you to make the floating palettes not children of the board, before accepting my commit. Because in the current way, there happen too many things, which we don't have fully under control. The floating palettes disappear or become irresponsive together with the board. But you never want that. And you also don't want such a complicated masking mechanism. My solution is that the palette becomes a Qt-Window temporarily, but leaving the parent in. That works currently. But I would not be surprised if in future Qt versions this cheat stops working.

### DIFF
--- a/src/core/UB.h
+++ b/src/core/UB.h
@@ -35,10 +35,6 @@
 
 #define UB_MAX_ZOOM 9
 
-#if defined(Q_OS_LINUX) || (defined(Q_OS_OSX) && (QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)))
-#define UB_REQUIRES_MASK_UPDATE 1
-#endif
-
 struct UBMimeType
 {
     enum Enum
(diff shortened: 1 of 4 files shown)


letsfindaway:
I stumbled across your PR while searching for ideas to improve Desktop mode for Wayland. I'm working with KDE, and think I have workarounds for most issues. But there are other problems when using Gnome: Interaction with applications is no longer possible after annotating something. And I thought perhaps your idea would help here.

I still have several problems with your PR. The palette does not stay in foreground, at least on KDE. Changing the palette from widget to window also has other implications on Wayland as on X11. The minimized palette always appeared in the center of the screen. With Wayland, you do not have the same control about window positions as on X11. So there is still some work to do. But I think the idea to make the glass pane fully transparent to mouse clicks when the Selector is used is interesting.

My question: would you like to help with the adaption to Wayland? The Qt version used is Qt 6. There are no plans to use OpenBoard Qt 5 on Wayland, as all fixes go to Qt 6 now. And what desktop environment are you using? Gnome, KDE, others?

Issue #873, opened 2024-01-17
Silent install for Windows isn't really working


SBAT73:
Hello
#### Expected behavior
I would like the installation to be completely silent.

#### Actual behavior
the 2 windows of the 2 vcredist are appearing on screen briefly

### Additional resources
*PDF, images, sounds... anything useful to reproduce the bug*

### Context

  - reproducibility : `systematic`
  - OpenBoard version :
      - bug happens with : `1.6.4` `1.7.0` 
      - didn't check :  other versions
  - OS : 
      - bug happens with : `Windows 10 & 11`
      - didn't check : `Linux` `macOS` 
  - Installation source : `OpenBoard download page` 
  
Thank you for your help

SBAT73:
it seems i've not declare properly. so i put my observations in the comment
Hello

### To Reproduce
<!---
I've try it on W10x64 & W11x64
with that commandline : 
OpenBoard_Installer_1.7.0.exe /VERYSILENT /NORESTART

In the installation log I see this : 

Filename: C:\Users\...\AppData\Local\Temp\is-D59NH.tmp\vcredist_2013.x64.exe
Parameters: /PASSIVE /VERYSILENT /SUPPRESSMSGBOXES /NORESTART
Process exit code: 0
-- Run entry --
Run as: Current user
Type: Exec
Filename: C:\Users\...\AppData\Local\Temp\is-D59NH.tmp\vcredist_2015_2019.x64.exe
 Parameters: /PASSIVE /VERYSILENT /SUPPRESSMSGBOXES /NORESTART
Process exit code: 0

So i've try those parameters directly on the vcredist executables. The "/PASSIVE" parameter doesn't work. A window is appearing on the screen.
But if I replace "/PASSIVE" by "/Q" or "/QUIET" it works


Thank you for your help

PR #899, opened 2024-02-22
Make silent installation fully silent

PatricCunhaforWork:
The current parameters for silent install of Visual C++ still display the following window (just an example, couldn't capture the exact one):

c1a4a99d-30d2-4a9c-9258-578d3a23ca30

Should fix the issue reported here: https://github.com/OpenBoard-org/OpenBoard/issues/873


SBAT73:
Thank you @PatricCunhaforWork.
Do you think a new release will fix this issue soon?

Issue #887, opened 2024-02-09
[Bug] Problema with window size when upgrading to 1.7.x - not an issue in 1.6.4


fbaldner:
### Describe the bug
When installing the new version 1.7.0 on my desktop, I get a limited window to draw one, that seems like a "slit" from the whole page. Like this one:

![OpenBoard_M6ahCIdicD](https://github.com/OpenBoard-org/OpenBoard/assets/35616989/46d5bf84-fc8e-4da8-8806-5ce885cd2ac4)

The gray area can't be drawn into.
If I go in the Documents part, the page preview is shown correctly (the page in question is in the third row, second column):

![OpenBoard_vlyvh4CNad](https://github.com/OpenBoard-org/OpenBoard/assets/35616989/49ab7863-0cbf-416a-a783-93d18a2b977f)

This also happens with version 1.7.1-rc240202

If I install version 1.6.4 the page is shown correctly:

![OpenBoard_7h7cbZG5jI](https://github.com/OpenBoard-org/OpenBoard/assets/35616989/cfbb80c1-1721-4290-8e1b-2972a5db5f6f)

Even if I now go back and install 1.7.x it will go back to that "slit" version of the page.

On my laptop, I have a weird zoom on my board:

![Openboard na v1 7 0](https://github.com/OpenBoard-org/OpenBoard/assets/35616989/ec7b7086-692d-4c0c-9a8c-fcf012a4b3e6)

In this version, on my laptop, if I zoom out, it only affects that page, not the whole document.

Whereas if I install 1.6.4 I get a regular view:

![Openboard na v1 6 4](https://github.com/OpenBoard-org/OpenBoard/assets/35616989/905f432c-d5b3-44bf-bc65-562447d4fa25)

### To Reproduce

1. Install version 1.7.0/1.7.1-rc240202
2. Open it and realizes either the "slit" (desktop) or the "weird zoom" (laptop)
3. Close OpenBoard
4. Install version 1.6.4
5. Open it and realizes that both are ok.

#### Expected behavior
When opening the board, I should have a proper view that allows me to write into my whole board, not just a specific area of the page or with the tools on top of the board.

#### Actual behavior
I get a cut view of the board in my desktop and a zoomed version in my laptop.

### Screenshots
Screenshots are above, in the problem description.

### Additional resources
*PDF, images, sounds... anything useful to reproduce the bug*

### Context

  - reproducibility : `systematic` `not systematic`
  - OpenBoard version :
[...]

Issue #846, opened 2023-12-12
Left and right side of imported PDF slides are cut off in the Control window


maehne:
In OpenBoard 1.7.0 (built using the AUR package on Arch Linux x86_64 and forced to use XWayland for rendering), the left and right of imported PDF slides are cut off in the Control window, when it is on a ultrawide screen (3440 x 1440 pixel). The slides are correctly diplayed in the Display window (1920 x 1200 pixel):

![grafik](https://github.com/OpenBoard-org/OpenBoard/assets/1174243/cf96a8e8-2651-48ad-8ed7-548a68be5b2c)

The PDF slides have been prepared using the LaTeX Beamer document class. OpenBoard sets in this case the page size always to "Custom size". If I change it to 16:9, only the current slide becomes smaller and covers only about 1/3 of the screen:

![grafik](https://github.com/OpenBoard-org/OpenBoard/assets/1174243/9941a13a-99fb-4073-8231-323ba368a4e7)

Personally, I would prefer that the whole slide is alway visible in the Control window - even if it does not fit into the Display window. A transparent frame or border line suffices in my eyes to indicate the visible screen area.

letsfindaway:
I think this is the same as #887. See also my PR #901.

But I think your idea of a transparent frame is appealing: Just give an indication what is visible on the display screen instead of the boxing as it is currently done. Will think about this and how it should behave. Should full interaction be possible also on the outer area?

maehne:
I agree that issue #887 is a duplicate of this one. IMHO, being able to interact with the slides (writing, erasing, etc.) with the slides outside the visible area is desirable, e.g, when you are approaching the boarder while writing. One would like to finish the annotation while speaking and then move that part into the visible area. Otherwise, the annotation would get cut off and it looks ugly if you then try to complete it in second attempt.

Besides a grayed-out area indicating the non-visible part, maybe some people would prefer a simple (dashed) rectangle as a (configurable) option.

PR #901, opened 2024-02-26
fix: scaling of display view

letsfindaway:
This PR fixes the display view scaling issue reported in #887 as follows:

- scaling shall be based on control view size instead of page size
- scaling has to take into account the system scale factor
- refactor: use `center()` function of `QRect` instead of own calculation

The problem (and also the resolution using this PR) can be verified in a setup where control and display screen have different aspect ratios.

### DIFF
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -249,7 +249,7 @@ void UBBoardController::setBoxing(QRect displayRect)
 {
     if (displayRect.isNull())
     {
-        mControlLayout->setContentsMargins(0, 0, 0, 0);
+        mControlView->setBoxing({});
         return;
     }
 
@@ -275,7 +275,7 @@ void UBBoardController::setBoxing(QRect displayRect)
             boxWidth = 0;
         }
 
-        mControlLayout->setContentsMargins(boxWidth, 0, boxWidth, 0);
+        mControlView->setBoxing({boxWidth, 0, boxWidth, 0});
     }
     else if (displayRatio > controlRatio)
     {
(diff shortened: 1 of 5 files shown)


letsfindaway:
According to the idea of @maehne in #846 I now indicate the visible area of the display screen on the control screen using a gray overlay. The area under the overlay can still be interacted with on the control screen. See here how it looks like:

grafik

The four blue dots indicate the corners of the display screen. For the two forms it is clear on the control screen which part is visible and what is hidden on the display screen.

maehne:
@letsfindaway: I really like your suggested improvement! The different aspect ratios become increasingly a problem, because 16:10 and 3:2 screens have become more popular over the past years while video projectors typically use 16:9 aspect ratio.

kaamui:
@letsfindaway Just found out that when moving the view, making any action centering the view (erase the board, or double cliking on the hand tool) make the grey borders not display correctly.

Moving the hand again update the whole visible region correctly.

letsfindaway:
> @letsfindaway Just found out that when moving the view, making any action centering the view (erase the board, or double cliking on the hand tool) make the grey borders not display correctly.
> 
> Moving the hand again update the whole visible region correctly.

Ok, I see. Looks as if the view is not properly updated in that case. I assume that this is a Qt problem. It does not happen for me with Qt6, but only with Qt5.

I could mitigate the problem with this crude hack:

```c++
void UBBoardController::centerOn(QPointF scenePoint)
{
    QPointF offset{1, 1};
    mControlView->centerOn(scenePoint - offset);
    mControlView->translate(offset.x(), offset.y());
    UBApplication::applicationController->adjustDisplayView();
}
```

It seems that `translate` is updating the foreground properly. So centering on some point beneath the target and then centering solves the issue also for Qt5.

letsfindaway:
The idea of setting the scrollbar position instead of the center (see https://github.com/OpenBoard-org/OpenBoard/issues/1009#issuecomment-2294790181) also fixes this issue. So I will further work on that direction and store/restore the scrollbar position of the view instead of the center point of the scene.

This solution is not a hack, but probably a better way to adjust the view position anyway.

PR #920, opened 2024-03-22
Update OpenBoard_pt_BR.ts

alexrsoares:
OpenBoard_pt_BR.ts has been updated


alexrsoares:
There is another version

PR #955, opened 2024-05-08
Shortcut configuration

letsfindaway:
This PR supersedes #686, which again is based on #460.


letsfindaway:
@kaamui: what do you think about providing this PR **_without_** the configuration dialog? It contains some remarkable improvements for shortcut handling anyway:

- Shortcuts for switching tools are active even when the tools palette is hidden. Currently the shortcuts no longer work when the palette is hidden using the top left button.
- You can temporarily switch tools by pressing and holding the keys. Consider you are drawing something. When now pressing and holding the shortcut keys for the eraser, you can switch to the eraser. When you then release the keys, you are back to the previous tool. This function might also be useful for the laser pointer.
- You can switch tools while drawing. You can e.g. start drawing using the pen, then while the mouse button is still down, you can press the shortcut for the line tool and continue with a straight line etc.
- I think also the option to use the shortcuts without the `Ctrl` modifier is quite useful.

kaamui:
Hi @letsfindaway,

I think the configuration dialog need improvement, so I agree we could benefit the improvements you mentioned in the next release and separate it from the configuration dialog planned for 1.8. I have some feedback to give you, I simply didn't find the time to write them down. 

Go for the two separate PRs !

src/core/UBPreferencesController.cpp
```diff
@@ -293,6 +423,21 @@ void UBPreferencesController::wire()
 
     // about tab
     connect(mPreferencesUI->checkSoftwareUpdateAtLaunchCheckBox, SIGNAL(clicked(bool)), settings->appEnableAutomaticSoftwareUpdates, SLOT(setBool(bool)));
+
+    // shortcut tab
+    connect(mPreferencesUI->shortcutTableView, SIGNAL(activated(const QModelIndex&)), this, SLOT(actionSelected(const QModelIndex&)));
```
kaamui: activated as a signal makes it weird, in UX terms. On Windows + Qt 6.9.1, for example, you can select each column independently, amd need to double-click for the recording input to become enabled. 

Should use clicked instead.

kaamui:
I changed Alt+F4 to Alt+F5 to quit OpenBoard, but as Alt+F4 is a "system shortcut", it is still working, so now when I try to change it to Alt+F4, it's closing the preferences instead of recording it 😄

letsfindaway:
> I changed Alt+F4 to Alt+F5 to quit OpenBoard, but as Alt+F4 is a "system shortcut", it is still working, so now when I try to change it to Alt+F4, it's closing the preferences instead of recording it 😄

Yes, the "system shortcuts" are not easily to be changed. There are more things which should be addressed:

- The tablet pen buttons don't work at all as intended. This needs more investigations, as I don't know how these button events arrive in Qt.
- For the input field, I also want to investigate in [QKeySequenceEdit](https://doc.qt.io/qt-6/qkeysequenceedit.html), which is especially made for this.
- And of course I better have to understand how "system shortcuts" arrive at the application and how they are handled. Anyway it might not a good idea to reassign them. But then it might be a good idea not to allow such keys or key combinations for custom shortcuts. I think the user complaining about the behavior of "Shift Insert" in a text field also is using a "system shortcut" we did not have in mind and which is not handled in OpenBoard.

kaamui (changes requested):
I forgot to submit my pending review

letsfindaway:
Here a new proposal sketch for the recording box:
```mermaid
block-beta
  columns 3
  Record
  space
  space
  Key_Sequence["Key sequence"]
  Mouse_Button["Mouse button"]
  Stylus_Button["Stylus button"]
  KeyInput["Ctrl+X"]
  MouseInput["Back"]
  StylusInput["Middle"]
  Accept
  Reset
  Abort
  classDef label fill:#ffffff,stroke:#fff;
  classDef input fill:#f8f8f8,stroke:#aaa,color:#999;
  class Key_Sequence label
  class Mouse_Button label
  class Stylus_Button label
  class KeyInput input
  class MouseInput input
  class StylusInput input
```
The idea:
- Have a "Record" button on the top. I would also add a red dot as icon to this button. The dot should blink when recording is active.
- Below three labels for the input fields in row 3
- The last row contains the "Accept", "Reset" and "Abort" buttons
- We will not dynamically change a button's meaning as it was before (Record changed to Accept)
- The blinking read dot is a good visual feedback for recording
- Buttons are enabled or disabled according to the status

**Edit:** Just tried out, looks horrible. 

Other idea is to do inline editing in the table. However I do not have a good idea how to allow the user to abort editing or to reset an entry to defaults.

kaamui:
I need to think more about it, and will come back here to give some feedback, when I have more time, not a priority at this point.

letsfindaway:
I rebased the branch and added a new commit addressing most of the discussed changes:

- It rearranges the buttons below the input fields
- red border around disabled table while recording
- red background to the Record button while recording
- double-click on an entry starts recording, as well as setting the focus to one of the input fields

I still did not remove the stylus button input. I want to discuss whether we should assign a fixed function to the second button and display this in the table.

Issue #860, opened 2024-01-05
Does this allow mutli-touch functionality on the tv's?[Question]


devinx1:

letsfindaway:
No, OpenBoard only supports single touch. Even things like "zoom with two fingers" are currently not supported.

kaamui:
Hi,

I'll just add that it is not planned at all because it is not a need expressed by our teachers in general.

ThomasLucky13:
Hi,
Our teachers need this functionality, so I added it. Here is a link to my [PR](https://github.com/OpenBoard-org/OpenBoard/pull/963).
It allows several people to draw at once on Board mode with tools Pen and Marker.

PR #963, opened 2024-05-15
support multi-touch

ThomasLucky13:
Add to OpenBoard multi touch drawing. It allows several people to draw at once on Board mode with tools Pen and Marker.

### DIFF
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -1927,10 +1927,7 @@ bool UBBoardView::hasSelectedParents(QGraphicsItem * item)
 
 void UBBoardView::releaseAllInputDevices()
 {
-    if (mMouseButtonIsPressed || mTabletStylusIsPressed || mPendingStylusReleaseEvent)
-    {
-        mMouseButtonIsPressed = false;
-        mTabletStylusIsPressed = false;
-        mPendingStylusReleaseEvent = false;
-    }
+    mMouseButtonIsPressed = false;
+    mTabletStylusIsPressed = false;
+    mPendingStylusReleaseEvent = false;
 }
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -1924,3 +1924,13 @@ bool UBBoardView::hasSelectedParents(QGraphicsItem * item)
         return false;
     return hasSelectedParents(item->parentItem());
 }
+
+void UBBoardView::releaseAllInputDevices()
+{
+    if (mMouseButtonIsPressed || mTabletStylusIsPressed || mPendingStylusReleaseEvent)
+    {
+        mMouseButtonIsPressed = false;
+        mTabletStylusIsPressed = false;
+        mPendingStylusReleaseEvent = false;
+    }
+}
--- a/src/domain/UBGraphicsScene.cpp
+++ b/src/domain/UBGraphicsScene.cpp
@@ -366,6 +366,8 @@ UBGraphicsScene::UBGraphicsScene(std::shared_ptr<UBDocumentProxy> document, bool
 //    connect(this, SIGNAL(selectionChanged()), this, SLOT(selectionChangedProcessing()));
     connect(UBApplication::undoStack.data(), SIGNAL(indexChanged(int)), this, SLOT(updateSelectionFrameWrapper(int)));
     connect(UBDrawingController::drawingController(), SIGNAL(stylusToolChanged(int,int)), this, SLOT(stylusToolChanged(int,int)));
+
+    installEventFilter(this);
 }
 
 UBGraphicsScene::~UBGraphicsScene()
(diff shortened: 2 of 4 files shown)


temaps:
https://github.com/OpenBoard-org/OpenBoard/issues/860

mikhailnov:
@ThomasLucky13 has been working very hard on this for quite a long time, and the resulting code looks tine and beautiful ;)

Vekhir (commented):
Thanks for your contribution, I believe that multi-touch is a useful feature to have.
I'm not that familiar with Qt's multitouch support, so I have some general remarks which you might use as an opportunity to expand on your design choices.

src/domain/UBGraphicsScene.cpp
```diff
@@ -3122,3 +3146,49 @@ void UBGraphicsScene::initStroke()
...
+        else if(distance < 4)
+            distance = 4;
+
+        UBBoardView* boardView = controlView();
+        QLineF line;
+        line.setP1(boardView->mapToScene(UBGeometryUtils::pointConstrainedInRect(lastPoint_m.toPoint(), boardView->rect())));
+        line.setP2(boardView->mapToScene(UBGeometryUtils::pointConstrainedInRect(endPoint_m.toPoint(), boardView->rect())));
+        if (!multiDrawLines.contains(line)) // to eliminate duplicates
+        {
+            multiDrawLines.append(line);
```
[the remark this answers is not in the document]
ThomasLucky13: This is important because there are so many duplicates that the page can take a long time to load

ThomasLucky13:
> Thanks for your contribution, I believe that multi-touch is a useful feature to have. I'm not that familiar with Qt's multitouch support, so I have some general remarks which you might use as an opportunity to expand on your design choices.

Thank you for code review. I was careless in some places and corrected them.

src/domain/UBGraphicsScene.cpp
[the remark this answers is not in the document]
Vekhir: If we are spending so much time on searching rather than appending, it might be worth using [`QSet`](https://doc.qt.io/qt-6/qset.html) instead of `QList`. `QSet` requires a bit more memory, but look-up is really fast.
Can you test it with `QSet` and report back whether it makes an appreciable difference in performance? Otherwise I'd just leave it as is.

src/board/UBBoardView.cpp
```diff
@@ -1924,3 +1924,13 @@ bool UBBoardView::hasSelectedParents(QGraphicsItem * item)
...
 }
+
+void UBBoardView::releaseAllInputDevices()
+{
+    if (mMouseButtonIsPressed || mTabletStylusIsPressed || mPendingStylusReleaseEvent)
+    {
+        mMouseButtonIsPressed = false;
+        mTabletStylusIsPressed = false;
+        mPendingStylusReleaseEvent = false;
+    }
```
letsfindaway, line 1935: I think the `if` is not really necessary here. The result is anyway to set all three variables to `false` and there is no performance benefit to test them first.

src/domain/UBGraphicsScene.cpp
```diff
@@ -2052,6 +2055,27 @@ bool UBGraphicsScene::isEmpty() const
     return mItemCount == 0;
 }
 
+bool UBGraphicsScene::eventFilter(QObject *watched, QEvent *event)
```
letsfindaway, line 2058: Is there a reason to use an `eventFilter` in the scene instead of overriding `event()` in the view? I would assume (but have not tested!) that the touch events arrive at the view and can be handled there. If this is the case, then I would also move the event handling to the view, while the multi-touch drawing code could be here in the scene.

What you're touching with your finger is a screen with a widget, which is the view. You don't "touch" the scene. The scene just holds all the graphics items.

src/domain/UBGraphicsScene.h
```diff
@@ -415,6 +417,8 @@ public slots:
         void hideMarkerCircle();
         void hidePenCircle();
         void DisposeMagnifierQWidgets();
+        void MultiTouchDrawing(QTouchEvent* event, UBStylusTool::Enum currentTool);
+        void MultiTouchEndDrawing();
```
letsfindaway, line 421: I most cases (well, most, there is a counter-example in the previous line, Sigh!) we start a method name with a lower case letter.

src/domain/UBGraphicsScene.cpp
```diff
@@ -3122,3 +3146,42 @@ void UBGraphicsScene::initStroke()
 {
     mCurrentStroke = new UBGraphicsStroke(shared_from_this());
 }
+
+void UBGraphicsScene::MultiTouchDrawing(QTouchEvent* event, UBStylusTool::Enum currentTool)
+{
+    QList <QTouchEvent::TouchPoint> touchPoints = event->touchPoints();
+    QPointF lastPoint_m, endPoint_m;
+    foreach (QTouchEvent::TouchPoint point, touchPoints)
```
letsfindaway, line 3154: Qt6 recommends to prefer the C++ range based `for` over the `foreach` keyword. See https://doc.qt.io/qt-6/foreach-keyword.html. It is also recommended to use a const reference in a case like this to avoid copying the point.

mikhailnov:
What is the status of this PR?

PR #987, opened 2024-05-26
Updated Greek Translation

mmpalasc:
I updated Greek Translation file. Should I do anything with the "unfinished" strings?

### DIFF
--- a/resources/i18n/OpenBoard_el.ts
+++ b/resources/i18n/OpenBoard_el.ts
@@ -12,11 +12,11 @@
     <name>BrowserWindow</name>
     <message>
         <source>Navigation</source>
-        <translation type="unfinished"></translation>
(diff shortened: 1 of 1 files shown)


kaamui:
Hi,

did you use linguist to do the translation ? If I remember well, validating a translation in the software is supposed to remove the unfinished attribute.

mmpalasc:
Hi,

I've never heard about linguist before. I translated the strings directly from the file. What should I do once I've downloaded linguist (I would like to use Qt linguist if possible)?

kaamui:
Just laucnh it, open the .ts file you want to modify. Then, in linguist, you'll see on each unfinished translation a question mark, select it and click on the little "check" icon to validate the string and go to next unfinished translation.

PR #990, opened 2024-05-27
Updated greek language

mmpalasc:
I updated the translation file using Qt Linguist, following @kaamui instructions in PR #987.
Is there such a way to translate the applications locales (they are in .js format) as well as html and pdf files from OpenBoard Website repository?
Also in some strings (which have singular and plural entries) I get the following warning:
"Translation does not contain the necessary %n/%ln place marker"
I translated those strings keeping the "%1" markers from the source. Should I do something else?

### DIFF
--- a/resources/i18n/OpenBoard_el.ts
+++ b/resources/i18n/OpenBoard_el.ts
@@ -12,102 +12,104 @@
     <name>BrowserWindow</name>
     <message>
         <source>Navigation</source>
-        <translation type="unfinished"></translation>
+        <translation>Πλοήγηση</translation>
     </message>
     <message>
         <source>Show downloads</source>
-        <translation type="unfinished"></translation>
+        <translation>Λήψεις</translation>
     </message>
 </context>
 <context>
     <name>CertificateErrorDialog</name>
     <message>
         <source>Dialog</source>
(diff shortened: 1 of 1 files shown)


kaamui:
> Is there such a way to translate the applications locales (they are in .js format) as well as html and pdf files from OpenBoard Website repository? 
Only in some of them. Look at `resources/library/applications` to which ones have locales or any working i18n system. 

Please note that an important work to do would be to translate the startupHints (located in `resources/startupHints`). Just copy/paste the `en` folder and translate every "user-visible" text inside the `.html` files

> Also in some strings (which have singular and plural entries) I get the following warning: "Translation does not contain the necessary %n/%ln place marker" 
Looks like a linguist bug. As long as you do the same as in english, there should not be any issue

mmpalasc:
Thanks for answering. startupHints are on my plans but I forgot to mention them in my last comment. As soon as I translate something, I'll create a pull request for each translation (hints, applications, etc.). By the way, I found some language.js files in "scripts" directory in Videopicker and Webbrowser wgt plugins. Should I translate them as well?

kaamui:
Yep, if you want the "show" button to be translated as well. But WebBrowser is deprecated and has been removed so you don't need to modify it ^^

PR #992, opened 2024-05-28
Paste in the middle of current view, not in the center

asivery:
When for example an image is pasted into the current document, it always lands in the center of it, regardless of where the user's view is located. This might be unintuitive for some users, who would expect the newly-pasted image to pop up on their screen, regardless of which part of the document they are looking at, and not have to scroll back to the document's center. This pull request fixes this issue.


kaamui:
Exact ! thank you !

PR #1001, opened 2024-06-06
Fix qt and c++ warnings

Vekhir:
Similar warnings are grouped into commits. If there are questions or suggestions, you may comment below, so we can discuss them. The solutions are based on the compiler recommendations.

Where changes cannot be made cross-compatible, they are version gated (mostly concerns Qt 6 vs Qt 5).  Some warnings remain that are a bit more involved (mainly around FFmpeg and OpenSSL/MD5).

Run on Arch Linux with Qt 6.7.1 and C++20 on GCC 14.1.1.

### DIFF
--- a/plugins/cffadaptor/src/UBCFFAdaptor.cpp
+++ b/plugins/cffadaptor/src/UBCFFAdaptor.cpp
@@ -631,7 +631,7 @@ QDomElement UBCFFAdaptor::UBToCFFConverter::parsePageset(const QStringList &page
 
     QDomElement svgPagesetElement = mDocumentToWrite->createElementNS(svgIWBNS,":"+ tIWBPageSet);
 
-    for (const auto &value : qAsConst(pageList)) {
+    for (const auto &value : std::as_const(pageList)) {
         svgPagesetElement.appendChild(value);
     }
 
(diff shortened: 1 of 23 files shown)


letsfindaway:
To check compatibility of this patch with various versions of ffmpeg and Qt,  I created a [project on the openSUSE build service](https://build.opensuse.org/project/show/home:letsfindaway:pr1001) to build the current `dev` branch together with this PR for Qt5 and Qt6 and a range of Arch, Debian, openSUSE and Ubuntu versions. **The results show no conflicts with any of these distribution versions**.

The `unresolvable` for the Qt6 build with Debian Unstable is not caused by this PR, but by the transition of Debian to `t64` packages mitigating the 32bit time problem for 32-bit builds. It is not relevant here for this check.

So my conclusion: **I support merging this PR, as it removes deprecations and improves compatibility with current and future Qt versions**.

letsfindaway (approved):
I also reviewed the changes manually and did not find any problems. I therefore approve this PR.

Vekhir:
@letsfindaway Thanks for having a look. While testing on my system, I've made sure to be as compatible as possible and keep the changes small. Great to see that the documentation for Qt and C++ is actually accurate :smile:

Issue #1004, opened 2024-06-06
[Bug] OpenBoard's user data directory depends on whether QT_QPA_PLATFORM is set to xcb or wayland


maehne:
### Describe the bug

While testing OpenBoard in a KDE Plasma Wayland session, I noticed that its user data directory depends on whether QT_QPA_PLATFORM is set to xcb or wayland.

### To Reproduce

1. Log into a KDE Plasma Wayland session and launch a shell
2. Launch OpenBoard using `QT_QPA_PLATFORM=wayland openboard`
    
```bash
$ QT_QPA_PLATFORM=wayland openboard
Persisting document; path is "~/.local/share/OpenBoard/document/OpenBoard Document 2024-06-06 15-18-39.461/metadata.rdf"
```
    
OpenBoard loads the documents from `~/.local/share/OpenBoard/document/` (notice the camelcase OpenBoard).

3. Launch OpenBoard using `QT_QPA_PLATFORM=xcb openboard` 
    
```
$ QT_QPA_PLATFORM=xcb openboard    
Qt: Session management error: None of the authentication protocols specified are supported
sAppSettings location:  "/etc/openboard/openboard.config"
Persisting document; path is "~/.local/share/openboard/document/OpenBoard Document 2024-06-06 15-20-23.932/metadata.rdf"
```
    
OpenBoard loads the documents from `~/.local/share/openboard/document/` (notice the all lowercase OpenBoard).

#### Expected behavior

OpenBoard should use the same user data directory independent of the setting of `QT_QPA_PLATFORM`.

#### Actual behavior

OpenBoard loads documents from `~/.local/share/OpenBoard/document/` when `QT_QPA_PLATFORM=wayland` and from `~/.local/share/openboard/document/` when `QT_QPA_PLATFORM=xcb`. In the latter case, I also get the following message at startup:

![grafik](https://github.com/OpenBoard-org/OpenBoard/assets/1174243/7795a552-63f0-4031-baf8-071896a3cbfb)

If one clicks on "Proceed", OpenBoard closes without any message. The message appears at every startup. The app can be used after clicking on "Cancel". Maybe, @taeyi1 observed something similar in issue #994 when his documents disappeared.

### Additional resources

I have noticed that OpenBoard uses three different data directories in `~/.local/share`:

```
$ ls -al ~/.local/share/?pen*
~/.local/share/openboard:
total 36
drwxr-xr-x  8 mht1 mht1 4096 Jun  6 15:18  .
[...]

Issue #951, opened 2024-05-07
Will you give a release build for ubuntu 24.04 too and in Appimage format too?


kumsumit:
I have Ubuntu 24.04, i think many have too.
Please consider to provide a release build for Ubuntu 24.04.
Please also consider to release in the AppImage format, because every Linux user doesn't have Ubuntu. 
AppImage format is useful for them.
Building from source is time consuming process and doesn't preferred by everyone
Thank you

letsfindaway:
In the meantime you might have a look at my [experimental builds](https://github.com/letsfindaway/OpenBoard/issues/161). Here I also build a release candidate of the upcoming OpenBoard 1.7.2 for Ubuntu 24.04.

sebojolais:
You could consider also the [FlatPak](https://flathub.org/apps/details/ch.openboard.OpenBoard).
It is mentioned on the [Download wiki](https://github.com/OpenBoard-org/OpenBoard/wiki/Downloads).

Michael24Git:
Hi,
I am using openboard always with Userdirectories for e.g. Data, Video, Audio etc. My productive environment is Ubuntu (Voyager) 22.04 and it works fine. I am using Userdirectories, because it is easier to sync just ONE Folder with my Nextcloud.

But I would like to switch to 24.04 in the near future and in my testing environment on another partition (Ubuntu Voyager 24.04.) I installed openboard in Ver. 1.7.2 RC and it seems to work, but it will totally ignore my settings in /etc/openboard.config with my directory-entries... like:
[App]
DataDirectory=/media/mypathto/Unterricht
UserImageDirectory=/media/mypathto/Bilder
UserVideoDirectory=/media/mypathto/Videos
UserAudioDirectory=/media/mypathto/Audio

and at
[podcast]
RecordingDirectory=/media/mypathto/Aufzeichnungen

The path /media/mypathto leads to another partition, where my data is stored.

So my question is, if you have canceled possibility for Userdirectories (hopefully not), or if you have any idea, what I can do with 1.7.2  on 24.04 and Userdirectories?
Best, Michael

letsfindaway:
For the location of configuration files see [Configuration at the Wiki](https://github.com/OpenBoard-org/OpenBoard/wiki/Configuration). The file is located in your home directory and is called `OpenBoardUser.config`.

[...]

PR #1010, opened 2024-06-09
fix: cast for UBApplication object during construction

letsfindaway:
- avoid to static_cast UBApplication object while constructor is still not completed
- fixes problem to determine userDirectory before ApplicationName and OrganizationName are defined

This PR fixes some strange problems we recently had with OpenBoard using the wrong user data directory. Especially it helps (or even solves?) the problems described in

- #1004 
- https://github.com/OpenBoard-org/OpenBoard/issues/951#issuecomment-2139682397

### DIFF
--- a/src/core/UBApplication.h
+++ b/src/core/UBApplication.h
@@ -80,7 +80,7 @@ class UBApplication : public SingleApplication
 
         static UBApplication* app()
         {
-            return static_cast<UBApplication*>qApp;
+            return dynamic_cast<UBApplication*>qApp;
         }
 
         static const QString mimeTypeUniboardDocument;


Vekhir:
From a cursory `grep` on `UBApplication::app()` it seems like the result is most of the time assumed to be non-nullptr without checking.
In the one case where it is checked, returning a nullptr would disable logging to a file, which doesn't look like it's the problem.
If you've checked and it solves #1004, great. From a theoretic point of view (I don't have this issue), this doesn't seem related.

Of course, the issue remains that `UBApplication::app()` is uninitialised memory. Since everything depends on the app, perhaps just aborting the program might be the safest route. The logging could get a check like `UBApplication::isInitialized` (using `dynamic_cast` like here) to make it clear that it could be called during initialisation.

letsfindaway:
> From a cursory `grep` on `UBApplication::app()` it seems like the result is most of the time assumed to be non-nullptr without checking. In the one case where it is checked, returning a nullptr would disable logging to a file, which doesn't look like it's the problem. If you've checked and it solves #1004, great. From a theoretic point of view (I don't have this issue), this doesn't seem related.

It is related in the following way:
grafik

- During startup, `UBApplication` is instantiated.
- As this inherits from `SingleApplication` and this again from `QApplication` these constructors are executed first.
- During initialization of `QApplication` a debug message is generated stating "Qt: Session management error: None of the authentication protocols specified are supported".
- The debug output is intercepted by `ub_message_output()`.
- This function tries to determine whether the message must be written to a log file and first calls `UBApplication::app()->isVerbose()`.
- But because `UBApplication`'s constructor is still not completed, the `mIsVerbose` variable is not initialized and this function returns an arbitrary value from uninitialized memory.
- In many cases this is non-null and interpreted as `true`. Therefore `UBSettings::userDataDirectory()` is called to determine the log file directory.
- As this is the first call to this function, `dataDirPath` is empty.
- I don't have the "App/DataDirectory" variable and therefore we determine the directory from `QStandardPaths::writableLocation(QStandardPaths::AppLocalDataLocation)`.
- As `setOrganizationName` and `setApplicationName` are still not executed, the name of the executable is used to construct the path. In my case this path ends with an all lowercase `openboard`.
- This path is then stored in `dataDirPath`, which is never changed later.
- Note that even the `getAppSettings()` uses a wrong file name for the settings for the same reason.
- Phew, this is a long chain of events and it took me quite some time to figure it out.

> Of course, the issue remains that `UBApplication::app()` is uninitialised memory. Since everything depends on the app, perhaps just aborting the program might be the safest route. The logging could get a check like `UBApplication::isInitialized` (using `dynamic_cast` like here) to make it clear that it could be called during initialisation.

I was thinking about better checks for `nullptr` but then did not do it for the following reasons:

- As soon as `qApp` is initialized, it **always** points to an instance of `UBApplication` and the `dynamic_cast` is **always** successful.
- While `qApp` is completely uninitialized, it contains `nullptr`. Here a `static_cast` and a `dynamic_cast` return the same value, i.e. also a `nullptr`. But this is only in the few lines from the beginning of `main()` to the construction of the `UBApplication` object and there is no usage of `qApp`.
- So the only difference in the return value is when `UBApplication::app()` is called **while** the object is constructed. This is never done directly in any code of OpenBoard. The only way I can think for such a call is the indirect invocation from `ub_message_output()` as explained above. And here the result of the cast is checked.

So I don't think that any further check is necessary.

Vekhir:
Thanks for detailing the causal relationship. I agree, this is the best course of action.

This shows how "set at first use" is flawed, but I can't think of anything better.

maehne (approved):
Looks good to me, as already stated in [issue #1004 ](https://github.com/OpenBoard-org/OpenBoard/issues/1004#issuecomment-2157602759). Thanks @letsfindaway for the detailed analysis of the problem's cause!

Issue #1007, opened 2024-06-07
[Feature Request]


MestniyPol:
Hi, why haven't you implemented auto-snap for the cursor when drawing lines or shapes yet? It's inconvenient to insert them manually through the menu in the application. I would like the application to straighten a curved line automatically when I draw it, or to auto-correct a circle when I draw one. Is it difficult to implement, and that's why it's not available?

Issue #871, opened 2024-01-16
Disable 45° line snap [Question]


freshmre:
45° snap in the "Draw Lines" is sometimes bothersome: requiring me to either use the ruler, draw a line and move it, or compromise precision.
The picture below shows the two nearest lines to B I can draw from A.
![image](https://github.com/OpenBoard-org/OpenBoard/assets/78566389/f64dca57-958e-406a-8fb9-d93333131f00)

Vekhir:
@freshmre You can use the ruler widget (`Applications` -> `Ruler`) to draw the line. When the line tool is selected, you can drag along the ruler to draw a line on its side. The ruler can be rotated in fine increments. Afterwards, the ruler can be deleted.

I don't think generally disabling the snap is a good idea since it removes the ability to have those precise angles which I think the tool was designed for.

letsfindaway:
fixed with PR #1013

PR #1013, opened 2024-06-15
Feat: snap to grid

letsfindaway:
This pull request adds a snap-to-grid function when drawing lines and moving tools or objects and resizing objects. In detail it provides the following functions:

- The `Shift` key is no longer used for multi-selection. Instead now only the `Ctrl` key works for that. Before, both keys were used for multi-selection.
- The `Shift` key is now used to activate snap-to-grid. The snap function is only active while this key is pressed. If `Shift` is not pressed, all draw, move and resize operations do not snap.
- This also means that the current 0°/45°/90° snap when drawing lines is no longer active when `Shift` is not pressed. This was requested by several people who wanted to draw "near horizontal" lines.

With the `Shift` key pressed, some points snap to the grid as follows:

- If the background is blank, then no snapping occurs.
- If the background is lined, then points snap to the closest line.
- If the background is crossed, points snap to the closest crossing.
- If intermediate lines are on, they also act as snap points.
- Additionally when drawing a line, the 0°/45°/90° snap is activated.

The individual tools and objects behave as follows when the `Shift` key is pressed and the tool or object is **moved**:

- The center of the **Axes** tool snaps.
- The Zero point of the upper scale on the **Ruler** snaps.
- The Zero point of the **Protractor** snaps.
- The Zero point of the **Triangle** snaps.
- The needle of the **compass** snaps.
- For **Objects**, all four corner points are snappy.

The individual tools and objects behave as follows when the `Shift` key is pressed and the tool or object is **resized**:

- When the **compass** is resized (radius is adjusted) and `Shift` is pressed, then the radius snaps so that the circle goes through the closest grid point.
- No snap when **Axes**, **Ruler**, **Protractor** or **Triangle** are resized.
- When **Objects** are resized, the resized side snaps to the grid.
- No snap for **Objects** if the lower right corner is resized. Here keeping the aspect ratio is more important.

When an object is in a rotated state, then the corners of the bounding rectangle of the rotated object define the points to snap.

For a **Line**, the snap points are not determined by the outline of the line, but by the actual line start and end points inside the outline. This works however only if the line is moved or resized in the same session where it was created. After a document was serialized and loaded again, the line property is lost and the line is handled like any other arbitrary drawing.

The implementation of the snap function is as follows:

- In `UBGraphicsScene`, functions are added to compute the snap vector, i.e. the amount by which we should move an object so that it snaps.
- Separate commits use these functions for the tools (Axes, Ruler, Protractor, Triangle and Compass).
- For drawing a line, the keyboard modifier is forwarded from the `UBBoardView` to the `UBGraphicsScene` and the snap functions are used.
- Moving objects using the Hand tool is handled in the `UBBoardView`.
- Moving and resizing objects with the pointer tool and using the Delegate frame is handled in the `UBGraphicsDelegateFrame`.

When at some later time the flexible background definitions are implemented (PR #888), then only the single `UBGraphicsScene::snap()` function must be adapted. All the other logic is not affected.

This PR is related to the following issues:

- #871 
- #1007

### DIFF
--- a/src/tools/UBGraphicsRuler.cpp
+++ b/src/tools/UBGraphicsRuler.cpp
@@ -47,6 +47,8 @@ UBGraphicsRuler::UBGraphicsRuler()
     : QGraphicsRectItem()
     , mResizing(false)
     , mRotating(false)
+    , mCursorRotationAngle(0)
+    , mItemRotationAngle(0)
 {
     setRect(sDefaultRect);
 
@@ -402,6 +404,10 @@ void UBGraphicsRuler::mousePressEvent(QGraphicsSceneMouseEvent *event)
     else if (rotateButtonRect().contains(event->pos()))
     {
         mRotating = true;
+        mCursorRotationAngle = 0;
+        QPointF topLeft = sceneTransform().map(boundingRect().topLeft());
+        QPointF topRight = sceneTransform().map(boundingRect().topRight());
+        mItemRotationAngle = QLineF(topLeft, topRight).angle();
         event->accept();
     }
     else
@@ -446,13 +452,24 @@ void UBGraphicsRuler::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
         {
             QLineF currentLine(rotationCenter(), event->pos());
             QLineF lastLine(rotationCenter(), event->lastPos());
-            rotateAroundCenter(currentLine.angleTo(lastLine));
+            mCursorRotationAngle = std::fmod(mCursorRotationAngle + lastLine.angleTo(currentLine), 360.);
+            qreal newAngle = mItemRotationAngle + mCursorRotationAngle;
 
-            //display current angle
+            if (event->modifiers().testFlag(Qt::ShiftModifier))
+            {
+                qreal step = UBSettings::settings()->rotationAngleStep->get().toReal();
+                newAngle = qRound(newAngle / step) * step;
+            }
+
+            newAngle = std::fmod(newAngle, 360.);
+
             QPointF topLeft = sceneTransform().map(boundingRect().topLeft());
             QPointF topRight = sceneTransform().map(boundingRect().topRight());
-            QLineF topLine(topLeft, topRight);
-            UBApplication::boardController->setCursorFromAngle(QString::number(topLine.angle(), 'f', 1));
+            qreal currentAngle = QLineF(topLeft, topRight).angle();
+            rotateAroundCenter(currentAngle - newAngle);
+
+            //display current angle
+            UBApplication::boardController->setCursorFromAngle(QString::number(newAngle, 'f', 1));
         }
 
         event->accept();
--- a/src/tools/UBGraphicsRuler.cpp
+++ b/src/tools/UBGraphicsRuler.cpp
@@ -420,6 +420,14 @@ void UBGraphicsRuler::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
     if (!mResizing && !mRotating)
     {
         QGraphicsItem::mouseMoveEvent(event);
+
+        // snap to grid
+        if (event->modifiers().testFlag(Qt::ShiftModifier)) {
+            // snap rotation center to grid
+            QPointF rotCenter = mapToScene(rotationCenter());
+            QPointF snapVector = scene()->snap(rotCenter);
+            setPos(pos() + snapVector);
+        }
     }
     else
     {
--- a/src/tools/UBGraphicsTriangle.cpp
+++ b/src/tools/UBGraphicsTriangle.cpp
@@ -555,7 +555,7 @@ void UBGraphicsTriangle::rotateAroundCenter(QTransform& transform, QPointF cente
 }
 
 
-QPointF    UBGraphicsTriangle::rotationCenter() const
+QPointF UBGraphicsTriangle::rotationCenter() const
 {
     return B1;
 }
@@ -807,6 +807,14 @@ void UBGraphicsTriangle::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
     if (!mResizing1 && !mResizing2 && !mRotating)
     {
         QGraphicsItem::mouseMoveEvent(event);
+
+        // snap to grid
+        if (event->modifiers().testFlag(Qt::ShiftModifier)) {
+            // snap rotation center to grid
+            QPointF rotCenter = mapToScene(rotationCenter());
+            QPointF snapVector = scene()->snap(rotCenter);
+            setPos(pos() + snapVector);
+        }
     }
     else
     {
--- a/src/tools/UBGraphicsTriangle.cpp
+++ b/src/tools/UBGraphicsTriangle.cpp
@@ -49,6 +49,8 @@ UBGraphicsTriangle::UBGraphicsTriangle()
     , mResizing1(false)
     , mResizing2(false)
     , mRotating(false)
+    , mCursorRotationAngle(0)
+    , mItemRotationAngle(0)
     , mShouldPaintInnerTriangle(true)
 
 {
@@ -787,6 +789,10 @@ void UBGraphicsTriangle::mousePressEvent(QGraphicsSceneMouseEvent *event)
     if(rotateRect().contains(event->pos()))
     {
         mRotating = true;
+        mCursorRotationAngle = 0;
+        QPointF topLeft = sceneTransform().map(boundingRect().topLeft());
+        QPointF topRight = sceneTransform().map(boundingRect().topRight());
+        mItemRotationAngle = QLineF(topLeft, topRight).angle();
         event->accept();
     }
     else
@@ -873,13 +879,24 @@ void UBGraphicsTriangle::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
         {
             QLineF currentLine(rotationCenter(), event->pos());
             QLineF lastLine(rotationCenter(), event->lastPos());
-            rotateAroundCenter(currentLine.angleTo(lastLine));
+            mCursorRotationAngle = std::fmod(mCursorRotationAngle + lastLine.angleTo(currentLine), 360.);
+            qreal newAngle = mItemRotationAngle + mCursorRotationAngle;
 
-            //display current angle
+            if (event->modifiers().testFlag(Qt::ShiftModifier))
+            {
+                qreal step = UBSettings::settings()->rotationAngleStep->get().toReal();
+                newAngle = qRound(newAngle / step) * step;
+            }
+
+            newAngle = std::fmod(newAngle, 360.);
+
             QPointF topLeft = sceneTransform().map(boundingRect().topLeft());
             QPointF topRight = sceneTransform().map(boundingRect().topRight());
-            QLineF topLine(topLeft, topRight);
-            UBApplication::boardController->setCursorFromAngle(QString::number(topLine.angle(), 'f', 1));
+            qreal currentAngle = QLineF(topLeft, topRight).angle();
+            rotateAroundCenter(currentAngle - newAngle);
+
+            //display current angle
+            UBApplication::boardController->setCursorFromAngle(QString::number(newAngle, 'f', 1));
         }
 
         //-----------------------------------------------//
--- a/resources/etc/OpenBoard.config
+++ b/resources/etc/OpenBoard.config
@@ -1,5 +1,4 @@
 [App]
-AngleTolerance=4
 HideCheckForSoftwareUpdate=false
 HideSwapDisplayScreens=true
 EnableAutomaticSoftwareUpdates=true
@@ -12,6 +11,7 @@ LastSessionPageIndex=0
 PageCacheSize=20
 PreferredLanguage=fr_CH
 ProductWebAddress=http://www.openboard.ch
+RotationAngleStep=5.
 RunInWindow=false
 SoftwareUpdateURL=http://www.openboard.ch/update.json
 StartMode=
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -299,21 +299,34 @@ void UBBoardController::setBoxing(QRect displayRect)
     }
 }
 
-void UBBoardController::setCursorFromAngle(QString angle)
+void UBBoardController::setCursorFromAngle(QString angle, const QPoint offset)
 {
         QWidget *controlViewport = controlView()->viewport();
 
         QSize cursorSize(45,30);
+        QSize bitmapSize = cursorSize;
+        int hotX = -1;
+        int hotY = -1;
 
-        QImage mask_img(cursorSize, QImage::Format_Mono);
+        if (!offset.isNull())
+        {
+            bitmapSize.setWidth(std::max(bitmapSize.width(), 2 * std::abs(offset.x())));
+            bitmapSize.setHeight(std::max(bitmapSize.height(), 2 * std::abs(offset.y())));
+            hotX = bitmapSize.width() / 2 - offset.x();
+            hotY = bitmapSize.height() / 2 - offset.y();
+        }
+
+        QSize origin = (bitmapSize - cursorSize) / 2;
+
+        QImage mask_img(bitmapSize, QImage::Format_Mono);
         mask_img.fill(0xff);
         QPainter mask_ptr(&mask_img);
         mask_ptr.setBrush( QBrush( QColor(0, 0, 0) ) );
-        mask_ptr.drawRoundedRect(0,0, cursorSize.width()-1, cursorSize.height()-1, 6, 6);
+        mask_ptr.drawRoundedRect(origin.width(), origin.height(), cursorSize.width()-1, cursorSize.height()-1, 6, 6);
         QBitmap bmpMask = QBitmap::fromImage(mask_img);
 
 
-        QPixmap pixCursor(cursorSize);
+        QPixmap pixCursor(bitmapSize);
         pixCursor.fill(QColor(Qt::white));
 
         QPainter painter(&pixCursor);
@@ -321,13 +334,13 @@ void UBBoardController::setCursorFromAngle(QString angle)
         painter.setRenderHints(QPainter::Antialiasing | QPainter::SmoothPixmapTransform);
         painter.setBrush(QBrush(Qt::white));
         painter.setPen(QPen(QColor(Qt::black)));
-        painter.drawRoundedRect(1,1,cursorSize.width()-2,cursorSize.height()-2,6,6);
+        painter.drawRoundedRect(origin.width() + 1, origin.height() + 1,cursorSize.width()-2,cursorSize.height()-2,6,6);
         painter.setFont(QFont("Arial", 10));
-        painter.drawText(1,1,cursorSize.width(),cursorSize.height(), Qt::AlignCenter, angle.append(QChar(176)));
+        painter.drawText(origin.width() + 1, origin.height() + 1,cursorSize.width(),cursorSize.height(), Qt::AlignCenter, angle.append(QChar(176)));
         painter.end();
 
         pixCursor.setMask(bmpMask);
-        controlViewport->setCursor(QCursor(pixCursor));
+        controlViewport->setCursor(QCursor(pixCursor, hotX, hotY));
 }
 
 
--- a/src/board/UBBoardController.h
+++ b/src/board/UBBoardController.h
@@ -165,7 +165,7 @@ class UBBoardController : public UBDocumentContainer
         void persistCurrentScene(bool isAnAutomaticBackup = false, bool forceImmediateSave = false);
         void showNewVersionAvailable(bool automatic, const UBVersion &installedVersion, const UBSoftwareUpdate &softwareUpdate);
         void setBoxing(QRect displayRect);
-        void setCursorFromAngle(QString angle);
+        void setCursorFromAngle(QString angle, const QPoint offset = {});
         void setToolbarTexts();
         static QUrl expandWidgetToTempDir(const QByteArray& pZipedData, const QString& pExtension = QString("wgt"));
 
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -391,13 +391,13 @@ void UBBoardView::tabletEvent (QTabletEvent * event)
     switch (event->type ()) {
     case QEvent::TabletPress: {
         mTabletStylusIsPressed = true;
-        scene()->inputDevicePress (scenePos, pressure);
+        scene()->inputDevicePress (scenePos, pressure, event->modifiers());
 
         break;
     }
     case QEvent::TabletMove: {
         if (mTabletStylusIsPressed)
-            scene ()->inputDeviceMove (scenePos, pressure);
+            scene ()->inputDeviceMove (scenePos, pressure, event->modifiers());
 
         acceptEvent = false; // rerouted to mouse move
 
@@ -409,7 +409,7 @@ void UBBoardView::tabletEvent (QTabletEvent * event)
         scene ()->setToolCursor (currentTool);
         setToolCursor (currentTool);
 
-        scene ()->inputDeviceRelease ();
+        scene ()->inputDeviceRelease (currentTool, event->modifiers());
 
         mPendingStylusReleaseEvent = false;
 
@@ -1146,7 +1146,7 @@ void UBBoardView::mousePressEvent (QMouseEvent *event)
                     connect(&mLongPressTimer, SIGNAL(timeout()), this, SLOT(longPressEvent()));
                     mLongPressTimer.start();
                 }
-                scene()->inputDevicePress(mapToScene(UBGeometryUtils::pointConstrainedInRect(event->pos(), rect())));
+                scene()->inputDevicePress(mapToScene(UBGeometryUtils::pointConstrainedInRect(event->pos(), rect())), 1., event->modifiers());
             }
             event->accept ();
         }
@@ -1311,7 +1311,7 @@ void UBBoardView::mouseMoveEvent (QMouseEvent *event)
 
     default:
         if (!mTabletStylusIsPressed && scene()) {
-            scene()->inputDeviceMove(mapToScene(UBGeometryUtils::pointConstrainedInRect(event->pos(), rect())) , mMouseButtonIsPressed);
+            scene()->inputDeviceMove(mapToScene(UBGeometryUtils::pointConstrainedInRect(event->pos(), rect())) , mMouseButtonIsPressed, event->modifiers());
         }
         event->accept ();
     }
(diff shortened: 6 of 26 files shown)


kaamui:
Very impressive work (and very well organized PR and commits !) ! I love it. I'll show it to the team asap.

src/tools/UBGraphicsRuler.cpp
```diff
@@ -466,6 +466,14 @@ void UBGraphicsRuler::mouseReleaseEvent(QGraphicsSceneMouseEvent *event)
     }
     else
     {
+        // snap to grid
```
kaamui, line 469: Shouldn't it be handled in the mouseMoveEvent instead ?

kaamui:
Hi @letsfindaway,

did you notice that when snapping an object, there is actually an intermediate position between two lines that is matching, making the moved object snapping at three different positions between two lines, where one would expect the object to only snap to the one of the two lines ?

Do you know what is causing this ?

https://github.com/user-attachments/assets/8f6b74ed-e3e1-44f3-9d82-eb68c587fa2f

letsfindaway:
> did you notice that when snapping an object, there is actually an intermediate position between two lines that is matching, making the moved object snapping at three different positions between two lines, where one would expect the object to only snap to the one of the two lines ?

This is because **all four sides of an object** are snappy. At one of the position, the bottom side snaps to the grid line, at the next the top side.

src/tools/UBGraphicsRuler.cpp
letsfindaway: That is not so easy here, because moving the tools is actually not handled by OpenBoard code, but left to Qt:

https://github.com/OpenBoard-org/OpenBoard/blob/1665e2e570e5e76c75d9e8a4b7e9a3f0c5071736/src/tools/UBGraphicsRuler.cpp#L420-L423

In such a case we would either have to modify much more code and implement the moving by ourselves, or modify the event, what I think is never a good idea. So I did the snapping here (and also for other tools) when releasing the mouse button.

T know that this leads to different behavior for different cases, but personally found that this does not hurt my experience.

kaamui:
> > did you notice that when snapping an object, there is actually an intermediate position between two lines that is matching, making the moved object snapping at three different positions between two lines, where one would expect the object to only snap to the one of the two lines ?
> 
> This is because **all four sides of an object** are snappy. At one of the position, the bottom side snaps to the grid line, at the next the top side.

OK. I expect it to be disturbing from a user's perspective, if left as-is. Do you think you could add a visual feedback of what snapped (a bluish or greyish line (or a corner) fade in and out where the object snapped) ? If too complicated, maybe we could detect what part of the selection frame was grabbed (top, left, right, bottom or inside) and react accordingly ?

Also, I noticed that if the user moves an object without the selection frame visible, the snap feature is not working.

src/tools/UBGraphicsRuler.cpp
kaamui: As I didn't implemented it and didn't know what to expect, I found it very counter-intuitive, and tried three of four times wondering what I was doing wrong with the tools, I only realized it was actually snapping after looking at the code. 

This snap feature is quite important imo, and I think it is very important that its behavior is harmonized, for moving as for rotating. I understand that it implies more code rework, but I think it is necessary, to guarantee a good user experience.

letsfindaway: > This snap feature is quite important imo, and I think it is very important that its behavior is harmonized, for moving as for rotating. I understand that it implies more code rework, but I think it is necessary, to guarantee a good user experience.

Ok, I understand. I think this affects all the C++ tools.

letsfindaway:
> OK. I expect it to be disturbing from a user's perspective, if left as-is. Do you think you could add a visual feedback of what snapped (a bluish or greyish line (or a corner) fade in and out where the object snapped) ? If too complicated, maybe we could detect what part of the selection frame was grabbed (top, left, right, bottom or inside) and react accordingly ?

When moving, you don't have to grab an object at a side. You can grab it in the middle. So another idea: make it dependent of the direction of the move: moving left snaps left edge etc.

> Also, I noticed that if the user moves an object without the selection frame visible, the snap feature is not working.

I see, this is only with the pointer tool. With the Play tool it always snaps. Will have a look at this.

letsfindaway:
> > * Additionally when drawing a line, the 0°/45°/90° snap is activated.
> 
> I think it could be more intuitive for the user that any object rotation behaves the same way, if Shift is pressed or not. Could you make the old logic (with angleTolerance) be active as long as a rotation is performed on any object, with Shift pressed ? What do you think ?

Of course line drawing is not rotation, but do I understand you right: You want to have a snap behavior also for rotations or arbitrary objects? But there is already something like this for rotation. It snaps at 30°, 45°, 60°, 90° etc independent of the Shift key.

Do you want that this snapping is also only active with Shift?

src/tools/UBGraphicsRuler.cpp
letsfindaway: It was much easier than expected and in fact, I can just move the code to `mouseMoveEvent`. For the compass, I also improved snapping of the pencil when resizing. It now also rotates the compass, so that you see exactly through which grid point the circle will go.

I have added the changes as a set of fixup commits and will merge them with the original commits before merging this PR. to keep the clean commit structure. Please tell me when you and your team are satisfied with the behavior.

letsfindaway:
I will rebase and merge the fixup commits into the corresponding original feature commits if you're satisfied with the behavior.

kaamui:
> When moving, you don't have to grab an object at a side. You can grab it in the middle. So another idea: make it dependent of the direction of the move: moving left snaps left edge etc.

For this I'm not sure, I'll present the feature to the team tomorrow, discuss this point with them and I'll keep you informed asap.

I think that making it dependent of the device move will create a behavior not immediately intuitive (Imagine you want to snap the bottom-left corner of an object with a corner of the grid located in the NE direction, you'll have to go beyond the desired position with the object, and then press Shift and come back in the SW direction).

You didn't answer to the idea of adding a visual effect/feedback when something snaps. What do you think of this idea ? I think it's the best because it lets all the possibilities that offers your first implementation, and increase the interactions/feedback with the user, which is a global lack of OpenBoard that need to be improved imo. I imagine it not too hard to implement but maybe it's here that I'm mistaking.

I'm currently pushing for more feedback, for depth handling for example (An improvement that I plan to propose to the team in this regard is the addition of a permanent indication of the depth of the object in the selection frame), and I actually like the first behavior you proposed, I just think we need to tell more to the user.

letsfindaway:
> You didn't answer to the idea of adding a visual effect/feedback when something snaps. What do you think of this idea ? I think it's the best because it lets all the possibilities that offers your first implementation, and increase the interactions/feedback with the user, which is a global lack of OpenBoard that need to be improved imo. I imagine it not too hard to implement but maybe it's here that I'm mistaking.

Ahh, I missed that. I think the visual feedback around an object is problematic when the object is not completely visible on the view. Then the feedback would also not be visible.

What about displaying the feedback near the cursor?

kaamui:
> Ahh, I missed that. I think the visual feedback around an object is problematic when the object is not completely visible on the view. Then the feedback would also not be visible.

Good point. But as the visual feedback would be very frequent during snapping, I think the user would know what's happening.

> What about displaying the feedback near the cursor?
Wow ! It could be very cool, and it would resolve the issue you pointed out. 

I saw that there is animation-related classes in Qt ([QPropertyAnimation](https://doc.qt.io/qt-6/qpropertyanimation.html)). So maybe it would be possible to have a blueish (I don't know why I imagine it blue, I may be influenced by diagrams.net ^^) or greyish line or angle appears and rapidly fades out, indicating what side or corner just snapped. 

Could it be around the cursor instead of near it (at the top-left of the cursor if top-left snap, at cursor's right if right side of the object snapped, etc) ?

letsfindaway:
I pushed an initial implementation of the snap indicator showing fading rectangles around the cursor. The visual appearance may be improved - proposals are welcome.

The indications are drawn in the new class `gui/UBSnapIndicator`, where also the default color is defined.

kaamui:
Hi @letsfindaway, 

First of all they like this feature. The main feedback is that the feature should not be disabled if no grid visible. it feels strange that Shift has no effect if no grid. 

I suggest that for empty backgrounds, the behavior is the same as with crossed background with its current values (so changing grid size also affects plain backgrounds).

Combined with your flexible ruler description, that we plan to merge too, but in a second time, I wonder how it should behave. If you have any idea, let me know.

letsfindaway:
> Hi @letsfindaway,
> 
> First of all they like this feature.

Glad to hear!!
 
> The main feedback is that the feature should not be disabled if no grid visible. it feels strange that Shift has no effect if no grid.

Hmm, I see. So let's see what to do.
 
> I suggest that for empty backgrounds, the behavior is the same as with crossed background with its current values (so changing grid size also affects plain backgrounds).

That would be very easy to implement. But I see some problems with the flexible ruler description. Which pattern should be the "default" used with a blank background? To keep your idea I wold propose to use a simple crossed background also there, because it is most logical for arranging items.
 
> Combined with your flexible ruler description, that we plan to merge too, but in a second time, I wonder how it should behave though. If you have any idea, let me know.

The idea is as follows:

- If the ruler description has only one line group, then it behaves like a lined background: items snap to the closest line.
- If the ruler description has two line groups, then items snap to the crossings of the lines of the two groups. 
- If the ruler description has three or more line groups, then we snap to all crossings of lines. But I think this is a rare use case. I currently only see the isometric grid: Here this rule means that we snap to the crossings.
- If the ruler description contains line groups with identical angle, then these are combined and handled as a single entity.

kaamui:
> The idea is as follows:
> [...]

No I meant what to do for the empty background, for example if crossed background is no longer on the list of templates. I did not look much at the code, so maybe it's not an issue because crossed background remains on the code no matter what background are available ?

letsfindaway:
> No I meant what to do for the empty background, for example if crossed background is no longer on the list of templates. I did not look much at the code, so maybe it's not an issue because crossed background remains on the code no matter what templates are available ?

We could define the empty background as a crossed background with invisible lines (alpha = 0).

kaamui:
Good idea !

kaamui:
The team and I also decided that the rotation behavior should be simplified. Pressing Shift while rotating anything (drawings, shapes, apps, tools, etc) on the board will apply a step by step rotation of x° (probably 5° by default). And probably this 5° step will be configurable in the preferences. Combined with your work, it should really improve user experience around objects manipulations

letsfindaway:
> The team and I also decided that the rotation behavior should be simplified. Pressing Shift while rotating anything (drawings, shapes, apps, tools, etc) on the board will apply a step by step rotation of x° (probably 5° by default). And probably this 5° step will be configurable in the preferences. Combined with your work, it should really improve user experience around objects transformations

Shall this also apply to drawing a straight line with the line tool?

kaamui:
> And do we need the angle indication as we have it when rotating an object in this case?

Yep, and I'm actually implementing that. I'll push it in a few minutes. I'm also changing the precision to 1 decimal, but I'm not finding a simple way to do so for the protractor. Don't hesitate to improve the next commit.

kaamui:
> hall this also apply to drawing a straight line with the line tool?

Yep

kaamui:
> @letsfindaway
[feat: snap angle while drawing line](https://github.com/OpenBoard-org/OpenBoard/pull/1013/commits/6265e726601ce4cc9f1733f4109972a60c2eac56)

Wow Thank you ! I just realized it could sound like if I was asking you to implement the feedback but I was going to do it myself, and just wanted to inform you ! But thank you !

Thankfully I decided to go on the triangle today, so we didn't work both on the same thing ^^.

letsfindaway:
> Thankfully I decided to go on the triangle today, so we didn't work both on the same thing ^^.

I'm going to work on snapping the rotation of a multiselection group. Will not do anything for the tools. But note that my [feat: stepwise rotation of items](https://github.com/OpenBoard-org/OpenBoard/pull/1013/commits/5c0e106d1768b3269b880f7e2c56bc9c24e129a2) commit includes an update of `UBSettings` for the variable to define the step angle. We should harmonize on that.

src/tools/UBGraphicsTriangle.cpp
```diff
@@ -555,9 +546,21 @@ void UBGraphicsTriangle::rotateAroundCenter(QTransform& transform, QPointF cente
 }
 
 
-QPointF    UBGraphicsTriangle::rotationCenter() const
+QPointF UBGraphicsTriangle::rotationCenter() const
 {
-    return B1;
+    QPointF center{B1};
```
kaamui, line 551: you're reverting 74570dc

letsfindaway: I started with the ruler: It is often used to measure some distance. Therefore I decided to use the zero of of upper scale as snap point. You can then use this to select the first point of the distance you want to measure. Then you rotate the ruler, until it goes through your second point. For this, it is essential that the zero point is the rotation center.

For the triangle I want to have the same behavior, as it also has a scale useful for measuring.

What are the arguments to use the edge of the triangle as rotation center instead?

kaamui: > Therefore I decided to use the zero of the upper scale as snap point

I'm OK with that, for the triangle too, but sill the edge of the triangle must remain the rotation center when it comes to rotating, as it was decided by the team. It's even more relevant now that the user can draw on each side of the triangle.

letsfindaway: > > Therefore I decided to use the zero of the upper scale as snap point
> 
> I'm OK with that, for the triangle too

Doesn't it then make sense to use the same corner used for rotation also as snap point? I still think it is not intuitive to use different points for snapping and rotating.

>  the edge of the triangle must remain the rotation center when it comes to rotating, as it was decided by the team

of course, if it is a team decision, then we have to stick to that. And for measuring distances we still have the ruler :)

kaamui: > Doesn't it then make sense to use the same corner used for rotation also as snap point? I still think it is not intuitive to use different points for snapping and rotating.

You're right. Let's use the same point for rotation as for snap points for every tool (with the edge being this point for the triangle, and the zero the one for the ruler)

> > the edge of the triangle must remain the rotation center when it comes to rotating, as it was decided by the team
> 
> of course, if it is a team decision, then we have to stick to that. And for measuring distances we still have the ruler :)

My commits, when not purely technical, are always based on what the team asked me to do :)

letsfindaway: I edited the commit for the triangle and reverted my changes of the rotation center. I think this was my last commit before our vacations. I will be back on 5.9.

kaamui: Thank you for all your work. Good vacations !

letsfindaway:
> I suggest that for empty backgrounds, the behavior is the same as with crossed background with its current values (so changing grid size also affects plain backgrounds).

Done.

letsfindaway:
> When at some later time the flexible background definitions are implemented (PR #888), then only the single `UBGraphicsScene::snap()` function must be adapted. All the other logic is not affected.

@kaamui If you want to try out how this will then behave, you could check out my branch [wip-snap-flex](https://github.com/letsfindaway/OpenBoard/tree/wip-snap-flex). This is a cherry-pick of the `feat-flexible-ruler-description` on top of `feat-snap-to-grid` including the necessary adaptations for the `snap()` function. For me it is quite impressive how e.g. drawing a line snaps on the isometric grid ;)

letsfindaway:
From my POV this PR is now ready for merging. Please let me know if you have still comments, questions and/or requests or if features are still missing or should be changed.

sebojolais:
I say thank you for this very useful feature and your impressive work.

One question : 
On crossed background, if I draw a horizontal line, the start point snaps to the cross grid but not the end point. Is it the expected behavior or there is something wrong in my side ?

sebojolais:
I would mentioned another issue with the `Seyes trick`:

- If the `Seyes trick` is enable
- The `Seyes background` is shown instead of the `lined background`.
- But the `snap to grid` works as it was the `lined background`
- So the user experience is affected. 

The solutions:
- The `flexible ruler PR` #888 will replaced the `Seyes trick` and it seems that it should work with `snap to grid` also (@letsfindaway) . So it will be one more argument for the `flexible ruler PR`.
- Or adjust the implementation of the `Snap to grid` to support the `Seyes trick` but it would be ugly...

letsfindaway:
> I say thank you for this very useful feature and your impressive work.
> 
> One question : On crossed background, if I draw a horizontal line, the start point snaps to the cross grid but not the end point. Is it the expected behavior or there is something wrong in my side ?
> 
> More details : I observed this issue for horizontal and vertical line. But the end point fits the crossed grid for diagonal lines.

Thanks for reporting! I looked into it: it is a race condition between "snap to grid" and "snap to angle". I have an idea how to solve that but I need some more time to investigate.

Currently this works as follows: When snapping the end point of a line, I consider two things:

- First the closest grid point.
- But also the end point of a line which is as long as the current line but is snapped to the nearest angle which is a multiple of 5°.
- Then the closest of those two is selected.

For most cases this works well, and you can easily snap to a grid point, but also if your cursor is far enough away from an grid point and close enough to the desired angle it snaps to the angle. Try it out for an angle of e.g. 35°!

But for angles of 0°, 45° etc the end point of the "snap-to-angle-line" is always closer than the grid point. Therefore snap-to-grid does not work for those cases.

My idea is to disable snap-to-angle if the line resulting from snap-to-grid would also satisfy the snap-to-angle condition, i.e. has an angle which is a multiple of 5°. Let me see how this is possible and how it behaves!

letsfindaway:
> I would mentioned another issue with the `Seyes hint`:
> 
> * If the `Seyes hint` is `enable`
> * The `Seyes background` is shown instead of the `lined background`.
> * But the `snap to grid` works as it was the `lined background`
> * So the user experience is affected.
> 
> The solutions:
> 
> * The `flexible ruler PR` [Flexible ruler description #888](https://github.com/OpenBoard-org/OpenBoard/pull/888) will replaced the `Seyes hint` and it seems that it should work with `snap to grid` also (@letsfindaway) . So it will be one more argument for the `flexible ruler PR`.
> * Or adjust the implementation of the `Snap to grid` to support the `Seyes hint` but it would be ugly...

Yes, you're right, and again thanks for reporting.

I can tell you that the flexible ruler description is also planned to be integrated, but most probably not for 1.7.2, as this is now in feature-freeze status, but for the next version.

I would recommend not to do any ugly workaround in the meantime but to tell people that this is a known limitation for that version which will be solved in the next.

sebojolais:
Thank you for your answers.

One more question:
On an `interactive whiteboard`, I used a **stylus** that can interact as a **left or right mouse button** only. Is there a way to benefit of the `snap to grid` feature in this usecase without any keyboard ?

letsfindaway:
> One more question:
> On an `interactive whiteboard`, I used a **stylus** that can interact as a **left or right mouse button** only. Is there a way to benefit of the `snap to grid` feature in this usecase without any keyboard ?

Currently: no. Do you have any idea how the user interaction could look like? Does the stylus have a button which could be used for that? Typically if the stylus has at least one button or switch, then this is used to switch to eraser. Other ideas like the introduction of a toggle button e.g. in the stylus toolbar to enable/disable snap mode have the drawback that you cannot e.g. start a line without snapping and then end it with snapping enabled.

And: I assume that this scenario is currently not a use case in Geneva, so any changes affecting the visible UI will most probably not be merged.

So the only idea I have: if the stylus has a button which is not used to switch to eraser mode, then we could investigate whether this button could behave like the `Shift` key.

sebojolais:
Thank you for your detailed answer. 
So for every `interactive whiteboard` with stylus setup, the feature is not available and there is a small regression because the previous vertical and horizontal line snap feature does not work now. Same thing for rotation snap.
@kaamui When you imagine that every previous snaps should now work with the Shift key, do you have this regression in mind ?
Thank you in advance for your answer.

sebojolais:
With an `interactive whiteboard` setup with a stylus, the main issue comes from that you have no `keyboard` in front of the `board`. 
But sometime you can move to the `keyboard`.
So **another solution** is possible, we can imagine to use the `Caplock` key (in complement to the `Shift` key):
- If the `cap-lock` is `on`, we are in `snap mode`
- If the `cap-lock` is `off`, we are in `normal mode`

The benefits are:
- no modification in the GUI
- and the Shift and Caplock keys work in a same way to type uppercase letters.

letsfindaway:
> With an `interactive whiteboard` setup with a stylus, the main issue comes from that you have no `keyboard` in front of the `board`. But sometime you can move to the `keyboard`. So **another solution** is possible, we can imagine to use the `Caplock` key (in complement to the `Shift` key):
> 
> * If the `cap-lock` is `on`, we are in `snap mode` 
> * If the `cap-lock` is `off`, we are in `normal mode`
> 
> The benefits are:
> 
> * no modification in the GUI 
> * and the Shift and Caplock keys work in a same way to type uppercase letters.

... but then you have to move between board and keyboard for that - not very comfortable.

Another idea: what about a long press. Start drawing your line, and don't move the pen for a second or so. This switches to snap mode for this line. Then draw the line, and it snaps to grid and angle. When lifting the pen again, the snap mode is disabled. 

We could first try this for line drawing. But it could also be applied for all the other snap operations. That would however need more considerations. Currently we have a lot of "mouse press" and "mouse move" event handlers for the operations. And you see from the number of my commits in this PR, that for each type of object or tool another handler needed to be adapted. This was not a problem when just considering the shift key. I needed an `if` anyway, and checking the `Shift` modifier is easy. However a long press handler needs a timer, and such code should not be replicated in many places.

However also here I would first ask @kaamui whether this is something he and his team could imagine.

kaamui:
I think a simpler - and probably better in terms of UX - would be to have a "magnet" button somewhere on the board to enable/disable snapping. 

> @kaamui When you imagine that every previous snaps should now work with the Shift key, do you have this regression in mind ?

No I didn't anticipate this issue, and I agree it's a regression for configurations with an interactive whiteboard. 

> Other ideas like the introduction of a toggle button e.g. in the stylus toolbar to enable/disable snap mode have the drawback that you cannot e.g. start a line without snapping and then end it with snapping enabled.

It's true but a far more acceptable drawback I think. With the button clearly visible, you know before starting your line if it will snap or not, and if you want it to snap or not, you click on it before starting your line.

kaamui:
And maybe @fadikdawoud can help with the enabled/disabled magnet icons ? 😃

kaamui:
> From my POV this PR is now ready for merging. Please let me know if you have still comments, questions and/or requests or if features are still missing or should be changed.

Our designer gave me the feedback that it would be better understood if the animation of the corner snapping was seen directly where it snaps, and that the use case of the snapping happening outside of the visible area is acceptable. 

I don't how if it is possible, but maybe we could use your work (around the mouse cursor) when not visible, and make it appear in the corner of the bounding box of the move item when visible ? Of course this change sounds more complicated and will be added in 1.7.3 if ever possible.

kaamui:
> And maybe @fadikdawoud can help with the enabled/disabled magnet icons ? 😃

The snap feature is a work done by @letsfindaway that gives the user the ability, when moving, resizing or rotating an object, to snap to the grid or not, by holding `Shift` when he wants to apply this "snap-to-grid" effect. @sebojolais noticed that it's not practical when you're on an interactive whiteboard with no direct access to a keyboard. 

So the good solution - imo - would be to have a button in the stylus toolbar (the one at the bottom of the board), or maybe right to this toolbar if we want to visually express that this button is not a tool like the other ones, that could be enabled/disabled before performing any transformation on an object of the scene. So it would imply to have to icons (a magnet is the common representation for this kind of features), one for when it is enabled, one for when it is not.

letsfindaway:
> > From my POV this PR is now ready for merging. Please let me know if you have still comments, questions and/or requests or if features are still missing or should be changed.
> 
> Our designer gave me the feedback that it would be better understood if the animation of the corner snapping was seen directly where it snaps, and that the use case of the snapping happening outside of the visible area is acceptable.
> 
> I don't if it is possible, but maybe we could use your work (around the mouse cursor) when not visible, and make it appear in the corner of the bounding box of the move item when visible ? Of course this change sounds more complicated and will be added in 1.7.3 if ever possible.

I think even this would not be too complicated, and could probably even simplify the code, because I then would move the appearing of the snap indicator to the `snap(QRectF)` function, where all necessary information is available.

letsfindaway:
> So the good solution - imo - would be to have a button in the stylus toolbar (the one at the bottom of the board), or maybe right to this toolbar if we want to visually express that this button is not a tool like the other ones, that could be enabled/disabled before performing any transformation on an object of the scene. So it would imply to have two icons (a magnet is the usual representation for this kind of features), one for when it is enabled, one for when it is not.

I definitely agree! And I would place the button to the right of the "Virtual Keyboard" button, which is also not a tool. Of course I can take the part of adding the code, and @fadikdawoud designs the icon. The other icon in the stylus palette are here: https://github.com/OpenBoard-org/OpenBoard/tree/dev/resources/images/stylusPalette . You see this are 42x42 px icons with transpatrent background. And I would prefer to have it as SVG, as this is more future proof when we better support HiRES displays.

kaamui:
> > > From my POV this PR is now ready for merging. Please let me know if you have still comments, questions and/or requests or if features are still missing or should be changed.
> > 
> > 
> > Our designer gave me the feedback that it would be better understood if the animation of the corner snapping was seen directly where it snaps, and that the use case of the snapping happening outside of the visible area is acceptable.
> > I don't if it is possible, but maybe we could use your work (around the mouse cursor) when not visible, and make it appear in the corner of the bounding box of the move item when visible ? Of course this change sounds more complicated and will be added in 1.7.3 if ever possible.
> 
> I think even this would not be too complicated, and could probably even simplify the code, because I then would move the appearing of the snap indicator to the `snap(QRectF)` function, where all necessary information is available.

If you think you can handle it during this month I can merge the modification in the 1.7.2

fadikdawoud:
> So the good solution - imo - would be to have a button in the stylus toolbar (the one at the bottom of the board), or maybe right to this toolbar if we want to visually express that this button is not a tool like the other ones, that could be enabled/disabled before performing any transformation on an object of the scene. So it would imply to have two icons (a magnet is the usual representation for this kind of features), one for when it is enabled, one for when it is not.

Thank you! That sounds like a solid approach. I'll start working on the icons right away and share the final versions here when they're ready.

> I definitely agree! And I would place the button to the right of the "Virtual Keyboard" button, which is also not a tool. Of course I can take the part of adding the code, and @fadikdawoud designs the icon. The other icon in the stylus palette are here: https://github.com/OpenBoard-org/OpenBoard/tree/dev/resources/images/stylusPalette . You see this are 42x42 px icons with transpatrent background. And I would prefer to have it as SVG, as this is more future proof when we better support HiRES displays.

I completely agree with SVG part! Placing the button to the right of the "Virtual Keyboard" seems like the best choice, as it aligns with the UI structure. I'd be happy to take on the task of designing the icon. Thanks for pointing me to the references—this will be really helpful. I'll make sure to follow the specifications and provide the icons as SVG files to ensure scalability and compatibility with Hi-Res displays.

> If you think you can handle it during this month I can merge the modification in the 1.7.2

I should be able to complete the designs within this month. Once everything is finalized, I’ll send the files over, and we can proceed from there. Thank you for your support.

letsfindaway:
I have now started #1100 and propose to continue the discussion there.

czyrichard:
> Additionally when drawing a line, the 0°/45°/90° snap is activated.

In version 1.7.2, it's more like a 0°/5°/10°/15°/20°/...90° snap. I wonder if there is a way to get the 0°/45°/90° snap back.

PR #1019, opened 2024-06-18
Remove MD5 hashing from UBFileSystemUtils

Vekhir:
The functions `MD5_Init`, `MD5_Update`, `MD5_Final` are deprecated in OpenSSL 3.0. This leads to warnings during compilation.

~~Use `EVP_Q_digest` instead.~~
**Edit:** Since the utility functions `md5` and `md5InHex` are not used anywhere, just remove them.

### DIFF
--- a/src/frameworks/UBFileSystemUtils.cpp
+++ b/src/frameworks/UBFileSystemUtils.cpp
@@ -41,7 +41,6 @@ THIRD_PARTY_WARNINGS_DISABLE
 #else
     #include "quazipfile.h"
 #endif
-#include <openssl/md5.h>
 THIRD_PARTY_WARNINGS_ENABLE
 
 #include "core/memcheck.h"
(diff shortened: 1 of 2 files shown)


letsfindaway:
I have another proposal: The MD5 functions can safely be removed from `UBFileSystemUtils` as nobody is using them. And if there would be some future use, then `QCryptographicHash` already provides all this functionality.

Vekhir:
True. I've checked out how the widgets sums [are calculated](https://github.com/OpenBoard-org/OpenBoard/blame/master/src/adaptors/UBWidgetUpgradeAdaptor.cpp#L236) and they use `QCryptographicHash`.

I've updated the commit to just remove those functions.

letsfindaway:
Please remove them also from the header file.

Vekhir:
Done

letsfindaway (approved):

PR #1023, opened 2024-06-20
Remove encrypted password storage

Vekhir:
OpenBoard saves the credentials for proxy connections and for uploading to YouTube. Since those need to be forwarded, they cannot be hashed, instead they are encrypted. This encryption is out of date and insecure.
The code has been imported from Open-Sankoré, i.e. it's over 13 years old. In the meantime, password managers have become widespread and easy-to-use, while being much more secure and better audited than the solution within OpenBoard. As such, they are better suited for storing the passwords, while OpenBoard will only ask and forward them without persistently storing them.

The proxy credentials need to be accessible from the `UBNetworkAccessManager`, so they are saved in a setting which is cleared at shutdown, i.e. no persistent storage.
The YouTube credentials are immediately forwarded, while the email can optionally be stored. The password is never saved.

In combination with #1019, this allows to get rid of the explicit OpenSSL dependency - it is of course still required implicitly via Qt.

Tested on Arch Linux with the community build and Qt6. While I have made sure that `qmake` should work too, I unfortunately cannot test that. I don't expect any differences with Qt5.


letsfindaway:
I have not tested, but I'm quite sure that the YouTube upload does not work in the current version of OpenBoard:
https://github.com/OpenBoard-org/OpenBoard/blob/753eadad57b92ce510b80f07775f20818d5f3201/src/podcast/youtube/UBYouTubePublisher.cpp#L107
This URL which is used for login produces a 404 error. In the end this means that the complete YouTube upload cannot work. 

Later the video is uploaded to the following URL:
https://github.com/OpenBoard-org/OpenBoard/blob/753eadad57b92ce510b80f07775f20818d5f3201/src/podcast/youtube/UBYouTubePublisher.cpp#L198
Here not even the server exists.

So the question is finally whether we should remove YouTube upload support completely - or fix it. I would vote for removal, as apparently nobody missed it.

Vekhir:
I thought as much, thanks for confirming. Web stuff has incredibly fast bitrot, wouldn't surprise me if uploading to YT didn't work already when it was imported...
Equally, the proxy stuff probably holds up better over time, but suffers the same standards changes.

In any case, removing YT support is better in a different PR, this one is already quite loaded for my taste. The removal of YT upload support entirely is probably also something that @kaamui could comment on.

Vekhir:
I made a local branch over at [Vekhir/OpenBoard#branch=remove-yt-upload-support](https://github.com/Vekhir/OpenBoard/tree/remove-yt-upload-support). It is based on this PR and removes upload support to YouTube entirely, seeing as it's both unmaintained and unused/unusable.

Issue #1046, opened 2024-08-14
[Building/Packaging] Remaining warnings by Qt6 and C++20


Vekhir:
**Describe the problem**
The PR #1001 removed many warnings that were cluttering the build logs while improving compatibility and potentially preempting future refactorings.
However, not all warnings were removed because they are trickier and might involve larger code changes. This issue shall provide a platform to collect and categorise these warnings and work towards a solution for them.
At the time of writing, 6 warnings remain.

**Versions**
Current dev (e9abc43)

**Build commands**
I'm using the `cmake` community build for Arch Linux, though `qmake` should be similarly affected. Full instructions: [PKGBUILD](https://github.com/bartoszek/AUR-openboard-git/blob/6d282596f0d6a6d4f7556aa3a3c882d092bfbcf7/PKGBUILD)

**Warnings**

<details>
    <summary>1. SingleApplication: QCryptographicHash::addData is deprecated</summary>

```
/build/openboard-git/src/OpenBoard/src/singleapplication/singleapplication_p.cpp: In member function ‘void SingleApplicationPrivate::genBlockServerName()’:
/build/openboard-git/src/OpenBoard/src/singleapplication/singleapplication_p.cpp:134:20: warning: ‘void QCryptographicHash::addData(const char*, qsizetype)’ is deprecated: Use the QByteArrayView overload instead [-Wdeprecated-declarations]
  134 |     appData.addData( "SingleApplication", 17 );
      |     ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/qt6/QtCore/QCryptographicHash:1,
                 from /build/openboard-git/src/OpenBoard/src/singleapplication/singleapplication_p.cpp:40:
/usr/include/qt6/QtCore/qcryptographichash.h:79:10: note: declared here
   79 |     void addData(const char *data, qsizetype length);
      |          ^~~~~~~
```
</details>
<details>
    <summary>2. UBPreferencesController.cpp: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20</summary>

```
/build/openboard-git/src/OpenBoard/src/core/UBPreferencesController.cpp: In lambda function:
/build/openboard-git/src/OpenBoard/src/core/UBPreferencesController.cpp:189:150: warning: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Wdeprecated]
[...]

PR #1047, opened 2024-08-14
refactor(UBPersistenceManager): Make createDocumentProxyStructure static

Vekhir:
The method `createDocumentProxyStructure` doesn't depend on the object, so can be made static. QtConcurrent::mapped requires the argument of the method to be const, since it creates a new list for the mapping.

The lambda function marks its arguments as const by default. Since the method is now static, the lambda can be replaced with a direct reference.

The removal of the lambda resolves warning 3 of #1046.


Vekhir:
Thanks

kaamui:
Thank you @Vekhir

Issue #827, opened 2023-11-19
Artifacts when moving/rotating objects


Vekhir:
## Issue
When moving or rotating an object "quickly", OpenBoard leaves artifacts visible. This is particularly prominent when zooming in, e.g. rotating the ruler at zoom level 2.1x leaves artifacts after almost any movement, making it really easy to test.
I was able to reproduce the issue with other widgets and self-made drawings. Lower zoom level (e.g. close to 1) usually means faster movement to create artifacts. One has to move the ruler rather quickly (only possible with a mouse, I think), but it's reproducible even at zoom level 0.2x (with the ruler widget at least).

## More information
The issue occurs both with Qt5 and Qt6, relevant versions below.
Interestingly, if I start a recording, then the artifacts stop occurring. The already existing ones stay until some other object comes near them, but no new artifacts are produced. This is true even after the recording is stopped - even if it is only a second long - and the podcast feature is disabled again (clicking "Openboard" > "Podcast" again). Was it just a moment ago basically impossible not to leave artifacts, it's now incredibly difficult to produce them. I _think_ I still got one small artifact afterwards, but they are essentially gone.
Restarting OpenBoard makes the artifacts occur again.

Is this something you are aware of, especially the behaviour with the podcast feature?

## System information
OS: Arch Linux
Kernel: 6.6.1.arch1-1
openboard-git: 1.6.4.r502.g1420d5f3-1 (current master)
Qt5: 5.15.11+kde+r146-1
Qt6: 6.6.0-3
ffmpeg: 2:6.0-13

Vekhir:
Based on my initial investigation above, the commit a504dbee0972353631bfc4f9b2e94f67aa4ad28c is at least partially responsible for the observed behaviour. The commit immediately preceding it (2f4f94aa965611936e6882fa30ecea3becefee48) does not show any artifacts even at 9.0x zoom level, while the offending commit above shows artifacts for any zoom level above 3x. For zooming levels below 3x, the commit does not show artifacts.

Vekhir:
The relevant commit touts a performance optimization in the paintEvent function whereby events obscured by the left palette are not drawn.
Further investigation revealed that closing the left panel also fixes above issue. No artifacts occur with the left sidepanel closed. Conversely, they reappear immediately after the left sidepanel is opened again.

Vekhir:
[...]

PR #1099, opened 2024-09-19
feat: display snap indicator at the respective corner

letsfindaway:
This PR addresses comment https://github.com/OpenBoard-org/OpenBoard/pull/1013#issuecomment-2360432207. The snap indicator is now shown at the point which caused the snapping. If this is outside of the visible area of the view, then the indicator is shown at the edge of the visible area.

- display snap indicator at the corner of the moved object which snapped
- limit indicator position by visible area of the view
- move appearing of the snap indicator to the snap(QRectF) function
- for rotated objects this makes it obvious that the corners of the bounding rectangle snap, not the corners of the object itself

Additionally this PR solves the problem with glitches when moving objects in a second commit.

- the already known problem with glitches when moving objects and the thumbnail palette is visible gets worse when moving objects with snapping
- this commit solves this problem by @Vekhir proposal to switch the view update mode depending on the current tool

See https://github.com/OpenBoard-org/OpenBoard/issues/827#issuecomment-1821556616

### DIFF
--- a/src/board/UBBoardView.cpp
+++ b/src/board/UBBoardView.cpp
@@ -857,12 +857,6 @@ void UBBoardView::handleItemMouseMove(QMouseEvent *event)
             movingItem->setPos(newPos);
 
             mLastPressedMousePos = scenePos + offset;
-
-            // display snap indicator
-            if (!offset.isNull())
-            {
-                updateSnapIndicator(corner);
-            }
         }
         else
         {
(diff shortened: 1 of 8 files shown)


letsfindaway:
If you want I can of course split this PR into two, because the two commits are not very much related.

kaamui:
No It's OK. But I'll have to change the design again. So for now, I'll simplify the behavior and we'll see how to improve it for 1.7.3.

PR #1104, opened 2024-09-21
Fixes Delete or Duplicate buttons on left panel

### LINKED ISSUE #1103: [Bug] Delete or Duplicate buttons on left panel do not work
In the `board view`, the `left panel` shows the pages of the current document. Each page has 2 buttons to `delete` or `duplicate` it.
These 2 buttons do not work in my side.

to reproduce :
- From the `dev` branch
- Ubuntu 22.04
- Qt 5.15.6

-> Solution and root cause:
Since this commit : https://github.com/OpenBoard-org/OpenBoard/commit/4730dc2cdfe1fae943a820cae201a60c3741d494
The thumbnail is filled by rescaled pixmaps.
So it seems that the buttons are also rescaled and the hard coded size of the icon (32) are not suitable.

The scale applied is :
```
1 / transform().m11()
```
So if we multiply the mouse coordinate by the inverse of this scale, it will work.

sebojolais:
Fixes #1103 issue.

### DIFF
--- a/src/gui/UBThumbnailWidget.cpp
+++ b/src/gui/UBThumbnailWidget.cpp
@@ -1031,7 +1031,7 @@ void UBDraggableThumbnailItem::moveDownPage()
 
 void UBDraggableThumbnailItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
 {
-    QPointF p = event->pos();
+    QPointF p = event->pos() * transform().m11();
 
     using namespace UBThumbnailUI;
 


letsfindaway:
Just as a description of the problem and the solution:

- During `paint()`, the buttons are scaled with `1 / transform().m11()` so that they keep their size independent of the item scale factor.
- When evaluating the mouse press event, we have to apply the inverse transformation to check whether we hit the button.

letsfindaway (approved):
Approved

Issue #1141, opened 2024-10-19
[Feature Request] Filter the documents


sebojolais:
**Is your feature request related to a problem? Please describe.**
My use-case of OB leads me to got lot of documents that are not always sorted in folders. So I have a very long list of document in the `document mode`. Some of them are rename with a name that make sense for me.
Currently I am not able to find quickly a document with an meaning name due to the list size. 

**Describe the solution you'd like**
If we add a `line edit` to **filter** the documents list, it will be very convenient.
We can image something like this:

https://github.com/user-attachments/assets/1de22a18-73e9-4b2a-aa66-f47e292cf94f

By this way, the list is shorter and I am able to find the right document.

**Describe alternatives you've considered**
I can imagine only one aspect of the feature that could be an issue : if the user types a string in the **filter** `line edit` and he forgets it and his action. So he will get a shorter list of documents that could leads him to think some documents are lost.

If we think that this risk could exist, we can set the `filter text` temporary, for example:
- At each time we quit the `document mode`, we clear the `filter lineedit`.
- After X seconds, we clear the `filter lineedit`.

**Additional context**
I did not find any topic about this feature on [the internal project page](https://github.com/orgs/OpenBoard-org/projects/1).

sebojolais:
I did not create a pull request before getting some feedback.
But the feature is available on [this branch](https://github.com/sebojolais/OpenBoard/tree/filteringDocuments).

letsfindaway:
Nice feature! And the implementation is so easy. I was just thinking whether a regular expression as you used it in the implementation is the right thing. Most users don't know much about RegExp, so a wildcard string could also be an option. But this are already the details.

letsfindaway:
> I can imagine only one aspect of the feature that could be an issue : if the user types a string in the **filter** `line edit` and he forgets it and his action. So he will get a shorter list of documents that could leads him to think some documents are lost.

At least we could set the property `clearButtonEnabled` on the `QLineEdit`. This is what most filters do. I would also support to clear the filter when leaving document mode, but I would not use a timer. I'm sure this would always trigger in the wrong moment.

sebojolais:
> Nice feature! And the implementation is so easy.

Yes, easy to implement it because everything was there. I had just to connect them together.

>  I was just thinking whether a regular expression as you used it in the implementation is the right thing. Most users don't know much about RegExp, so a wildcard string could also be an option. But this are already the details.

The Qt API to filter: 
[...]

PR #1142, opened 2024-10-20
Filtering documents

sebojolais:
Implements this feature request: #1141 

Add a `lineedit` to filter the documents list.


src/document/UBSortFilterProxyModel.h
```diff
@@ -10,6 +10,6 @@ class UBSortFilterProxyModel : public QSortFilterProxyModel
     UBSortFilterProxyModel();
 
     bool lessThan(const QModelIndex &left, const QModelIndex &right) const;
+    bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const;
```
letsfindaway, line 13: In the base class, `filterAcceptsRow()` is protected. A derived class should not change visibility of the function. I therefore propose to add `protected:`. And as it is the same with `lessThan()`, I would add this before line 12 and fix it for this function, too.

And as both functions are overriding a base class function, I would also add `override` to both.

And an unrelated side remark: This header contains a completely unnecessary `#include "core/UBPersistenceManager.h"`. I know, this has nothing to do with this PR, but as we'er already touching this header, shouldn`t we also clean up that? Typically I do not add unrelated cleanups to a PR, but here it is so obvious...

src/document/UBSortFilterProxyModel.cpp
```diff
@@ -23,3 +23,19 @@ bool UBSortFilterProxyModel::lessThan(const QModelIndex &left, const QModelIndex
 
     return QSortFilterProxyModel::lessThan(left, right);
 }
+
+bool UBSortFilterProxyModel::filterAcceptsRow(int sourceRow,
+                                              const QModelIndex &sourceParent) const
+{
+    QModelIndex index0 = sourceModel()->index(sourceRow, 0, sourceParent);
+    if(UBPersistenceManager::persistenceManager()->mDocumentTreeStructureModel->isCatalog(index0))
```
letsfindaway, line 31: Instead of using the `UBPersistenceManager` to get the model you should better use the same code as in `lessThan()`, cast the result of `sourceModel()` and check for `nullptr`. This avoids the implicit assumption that the model, where this filter works upon, is exactly that `mDocumentTreeStructureModel` managed by the `UBPersistenceManager`.

See also below for removing the include of the `UBPersistenceManager` in the header.

src/document/UBDocumentController.cpp
```diff
@@ -2316,6 +2316,7 @@ void UBDocumentController::setupViews()
 
         connect(mDocumentUI->sortKind, SIGNAL(activated(int)), this, SLOT(onSortKindChanged(int)));
         connect(mDocumentUI->sortOrder, SIGNAL(toggled(bool)), this, SLOT(onSortOrderChanged(bool)));
+        connect(mDocumentUI->filterText, SIGNAL(textChanged(const QString&)), this, SLOT(onFilterTextChanged(const QString&)));
```
letsfindaway, line 2319: I prefer the new syntax for making connection using function pointers and would like to use it in any new code. This already provides compile time checks for signals and slots. It also makes it easier to find references to these signals and slots in QtCreator.

letsfindaway (changes requested):
Thanks for your PR! I have a few comments. Please have a look at them.

src/document/UBSortFilterProxyModel.cpp
```diff
@@ -23,3 +23,19 @@ bool UBSortFilterProxyModel::lessThan(const QModelIndex &left, const QModelIndex
...
+    {
+        // Always show the catalog folders
+        return true;
+    }
+    else
+    {
+        // Filter the documents
+        return (sourceModel()->data(index0).toString().contains(filterRegularExpression()));
```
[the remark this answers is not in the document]
sebojolais: Thank you for your detailed review.
This topic is done.

src/document/UBSortFilterProxyModel.h
sebojolais: Ok, it is done.
Me too, I am not comfortable to change code with no relation to the PR. But it is done.
When I remove `#include "core/UBPersistenceManager.h"`, it leads me to adjust other includes on unrelated files.

src/document/UBSortFilterProxyModel.cpp
sebojolais: Done

src/document/UBDocumentController.cpp
sebojolais: Done

src/document/UBSortFilterProxyModel.h
letsfindaway: > When I remove `#include "core/UBPersistenceManager.h"`, it leads me to adjust other includes on unrelated files.

But on each of those occurrences you fixed a "bad" include. Those files did only compile because they were relying on the side effect from the `UBSortFilterProxyModel.h`.

letsfindaway (approved):
Approving this PR.

sebojolais:
The last commit implements the deletion of the filter linedit when we switch to the document mode, in order to reduce the risk explained [here](https://github.com/OpenBoard-org/OpenBoard/issues/1141#issue-2599623739):

> If the user types a string in the filter line edit and he forgets it and his action. So he will get a shorter list of documents that could leads him to think some documents are lost.

For information, it is tested on Linux Ubuntu 22.04 with Qt5 and Qt 6.7 with success.

letsfindaway:
Your PR has conflicts. Please rebase your branch to adapt to the current `dev`.

src/document/UBDocumentController.cpp
```diff
@@ -2392,6 +2393,12 @@ void UBDocumentController::onSortKindChanged(int index)
     UBSettings::settings()->documentSortKind->setInt(index);
 }
 
+void UBDocumentController::onFilterTextChanged(const QString& filter)
+{
+    mSortFilterProxyModel->setFilterCaseSensitivity(Qt::CaseSensitivity::CaseInsensitive);
+    mSortFilterProxyModel->setFilterRegularExpression(filter);
```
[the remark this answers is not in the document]
sebojolais: > Another questioning behavior is that we can perform additive actions with a filter set. Thus, we can create a document that will not be visible, same for a folder, or an import.
> 
> I think we should probably clear the filter when such operations are performed

I did a naive implementation of your description. It works fine. But this line : 
`    mDocumentUI->filterText->clear();
`
at each beginning of functions is quite ugly and not very meaningful for future review.

kaamui:
Thank you @sebojolais !

kaamui:
After further testing, I realize that finding a folder but not seeing its documents is an issue.

I'll push a modification for this

PR #1158, opened 2024-11-06
fix: import file only once to document

### LINKED ISSUE #1156: [Bug] Import PDF -> First page duplicated
### Describe the bug
When I import a PDF on the board, the first page is duplicated and paste at the end.

### To Reproduce
- Open OpenBoard
- Create an new document
- Go to board view
- Drag and drop the attached PDF file [Untitled 2.pdf](https://github.com/user-attachments/files/17631561/Untitled.2.pdf)
- The 2 pages of the PDF are added as expected
- But there is a third page that is the first page of the PDF with a different scale.

This behavior happens for every kind of PDF and OB documents.

#### To not reproduce
- Open OpenBoard
- Go to document view
- Import the attached PDF file [Untitled 2.pdf](https://github.com/user-attachments/files/17631561/Untitled.2.pdf)
- The new document has only 2 pages as expected

#### Expected behavior
We expect to add the number of pages of the PDF, no more.

#### Actual behavior
The Drag and Drop on the board addes the first page at the end

### Context
  - OpenBoard version : last dev branch 78ee9453ce924e36414808f40b52666bc51e5aa5
  - OS : 
      - bug happens with : `Linux` Ubuntu 22.04
      - didn't check : `Windows` `macOS` 
  - If you built OpenBoard from source, please provide the branch you're working on : `dev`
  
### Additional context
OS detailed version is MANDATORY (if not provided, the issue will be closed as incomplete). 
KDE Neon 6.2 (Ubuntu 22.04 based)

letsfindaway:
This PR fixes #1156 where the first page of a PDF document was repeated at the end when importing the PDF by dropping it to the board.

- when importing a file to the board, the file was offered to all import adapters
- if multiple import adapters support the same file extension, then the file was imported multiple times
- this was especially true for pdf documents, where in some environments also the image adapter supported rendering a pdf document (preview image of first page)
- this commit leaves the loop iterating over the adapters once one of them has successfully imported the file

@sebojolais: could you test this PR whether it fixes the problem for you?

### DIFF
--- a/src/core/UBDocumentManager.cpp
+++ b/src/core/UBDocumentManager.cpp
@@ -257,12 +257,12 @@ int UBDocumentManager::addFilesToDocument(std::shared_ptr<UBDocumentProxy> docum
         QFile file(fileName);
         QFileInfo fileInfo(file);
 
+        UBApplication::setDisabled(true);
+
         foreach (UBImportAdaptor *adaptor, mImportAdaptors)
         {
             if (adaptor->supportedExtentions().lastIndexOf(fileInfo.suffix().toLower()) != -1)
             {
-                UBApplication::setDisabled(true);
-
                 if (adaptor->isDocumentBased())
                 {
                     //issue 1629 - NNE - 20131212 : Resolve a segfault, but for .ubx, actually
@@ -270,7 +270,10 @@ int UBDocumentManager::addFilesToDocument(std::shared_ptr<UBDocumentProxy> docum
                     UBDocumentBasedImportAdaptor* importAdaptor = dynamic_cast<UBDocumentBasedImportAdaptor*>(adaptor);
 
                     if (importAdaptor && importAdaptor->addFileToDocument(document, file))
+                    {
                         nImportedDocuments++;
+                        break;
+                    }
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


sebojolais:
Hi,
Thank you for your PR.
It fixes the issue #1156 I observed in my side.
tested with Qt 5.15 and 6.7

sebojolais:
One question about this PR:

It make sense to let only one adapter to import the document.
But how we can be sure is the best adapter for this format ?
In other words, with a example:
Currently, the adapters list is : 
image

The PR works fine because the `PDF adapter` is before the `Image adapter` in the list.
If it was not the case the PR should not fixes the #1156 issue.
But it will be always the case because the order of the adapters is defined in the constructor:
https://github.com/OpenBoard-org/OpenBoard/blob/3b38085e95d9621d92afae079946fc13442cb6c2/src/core/UBDocumentManager.cpp#L94-L103

So I try to answer to my question: But how we can be sure is the best adapter for this format ?
We can not be sure, but the double adapters import happens only for PDF in a specific environment. And in this case the good PDF adapter will be always used first.

letsfindaway:
> So I try to answer to my question: But how we can be sure is the best adapter for this format ?
> We can not be sure, but the double adapters import happens only for PDF in a specific environment. And in this case the good PDF adapter will be always used first.

Yes, exactly. This is a very good question, and I also asked it to myself. To make the answer a little bit more generic:

To find the right adapter when multiple adapters would offer to handle a file format we can have two approaches:
- Some kind of voting, where each adapter expresses a "suitability" and we choose the highest.
- Or a fixed sequence, which orders the preferred adapter first.

The current code guarantees a fixed order, so the second approach works well. If we care to order more specific adapters before more generic ones, then this solves the problem.

And as you said, PDF is currently the only one where this occurs. And the image adapter is the only one where the list of supported file formats can vary. All others return a fixed string list.

kaamui:
yes, please

letsfindaway:
> yes, please
done

Issue #1164, opened 2024-11-17
[Bug] When RightLibPaletteBoardModeIsCollapsed=true the main Window stays invisible.


qqqqqqqqq9:
### Describe the bug

I use OpenBoard on Plasma/Wayland. For me the main window stay invisible. It shows in the pager and the window list.

### To Reproduce

1. Start OpenBoard.  
2. Close the right tab.
3. Close OpenBoard.
4. Start OpenBoard.

#### Expected behavior
OpenBoard should start.

#### Actual behavior
OpenBoard started but the window was invisible.

Workaround: 
Open the config and set

RightLibPaletteBoardModeIsCollapsed=true

### Context

  - reproducibility : `systematic`
  - OpenBoard version :
      - bug happens with : `1.7.1`
            - didn't check : `1.6.4` `1.7.0` 
  - OS : 
      - bug happens with :  `Linux` 
       - didn't check : `Windows` `macOS` 
I built from Source on opensuse using
 https://aur.archlinux.org/packages/openboard
as template. (qt6)

### Additional context

    - Linux : `openSuSE Tumbleweed `

letsfindaway:
Did you know that there is an official package for OpenBoard 1.7.1 in the openSUSE repositories?
https://software.opensuse.org/package/OpenBoard?search_term=Openboard

And I could not reproduce the problem here. Could you try it with the official package? Have you any other information about the problem?

qqqqqqqqq9:
Hi, 
yes I know about the official package but I wanted to try a qt6-version.

When I run from a terminal it spit's out the following:

libva info: VA-API version 1.22.0
libva info: Trying to open /usr/lib64/dri/radeonsi_drv_video.so
libva info: Found init function __vaDriverInit_1_22
libva info: va_openDriver() returns 0
Persisting document; path is "... edited custom path .../OpenBoard Document 2024-11-17 20-40-15.767/metadata.rdf"
eglSwapBuffers failed with 0x300d, surface: 0x0

qqqqqqqqq9:
The official version works find.

letsfindaway:
> eglSwapBuffers failed with 0x300d, surface: 0x0

When searching for this error message on the web I find quite some hits, e.g.

- https://bugreports.qt.io/browse/QTBUG-129403
- https://forum.qt.io/topic/158667/invisible-window-created-with-eglswapbuffers-error-0x300d
- https://bugs.kde.org/show_bug.cgi?id=455913
[...]

Issue #1041, opened 2024-08-01
Crashes using multi-displays - Flatpaks


fvelsg:
### To Reproduce

1. Go to the main screen
2. Click Openboard: ![image](https://github.com/user-attachments/assets/74e947a1-afe9-4882-a3dc-9692a53d225a)
3. Click on Preferences
4. Click on Display
5. Go to Multi Display
6. Click on `Use all available displays`
7. It will crash

#### Expected behavior
It should give me the option to change the main screen and the mirror screen.

#### Actual behavior
It just crashes right after I click the option

### Screenshots
![Screenshot from 2024-08-01 08-33-13](https://github.com/user-attachments/assets/72877a13-270b-4fd0-b498-a461993b2d9a)
![Screenshot from 2024-08-01 08-35-13](https://github.com/user-attachments/assets/09a2ff28-b8f3-41ac-833e-d49c9a8bbff1)
![image](https://github.com/user-attachments/assets/aada46e3-0a5a-4d01-9e36-a186050d2f7f)

### Context

  - reproducibility : `systematic`
  - OpenBoard version :
      - bug happens with : `1.7.1`
      - bug doesn't happen with : .deb version
      - didn't check : `1.6.4` `1.7.0` 
  - OS : 
      - bug happens with :  `Linux` 
      - didn't check : `Windows` `macOS` 
  - Installation source :`flatpak` 
  
### Additional context

- OS(es) detailed version (for the ones you reproduced the bug on) 
    - Linux : `Description:	Fedora 40 Workstation`
- number and resolution of monitors : ` 2` `1680x1050` `1920x1080`

letsfindaway:
Are you using Wayland? Then this is a known bug. A workaround is to set `QT_QPA_PLATFORM=xcb` before launching OpenBoard or to switch to X11.

Issue #843, opened 2023-12-10
OpenBoard 1.7.0 on Wayland: Segmentation fault when clicking into the "List of screens used for Control, Display and Previous pages" entry field


maehne:
I installed OpenBoard 1.7.0 on an up-to-date Endeavour OS (based on Arch Linux, x86_64 processor platform) installation using the [AUR PKGBUILD](https://aur.archlinux.org/packages/openboard). The AUR package currently builds OpenBoard agains Qt 5.15.11. A segmentation fault happens every time, when clicking with the mouse into the "List of screens used for Control, Display and Previous pages" entry field of the Preferences window. The backtrace is:

```
Thread 1 "openboard" received signal SIGSEGV, Segmentation fault.
0x00007fffec8a051b in QMetaObject::cast (
    this=this@entry=0x7ffff56856a0 <QWidgetWindow::staticMetaObject>, 
    obj=obj@entry=0x555557259ab0) at kernel/qmetaobject.cpp:389
Downloading source file /usr/src/debug/qt5-base/qtbase/src/corelib/kernel/qmetaobject.cpp
389         return (obj && obj->metaObject()->inherits(this)) ? obj : nullptr;  
(gdb) bt
#0  0x00007fffec8a051b in QMetaObject::cast(QObject const*) const
    (this=this@entry=0x7ffff56856a0 <QWidgetWindow::staticMetaObject>, obj=obj@entry=0x555557259ab0) at kernel/qmetaobject.cpp:389
#1  0x00007fffec8a055a in QMetaObject::cast(QObject*) const
    (this=this@entry=0x7ffff56856a0 <QWidgetWindow::staticMetaObject>, obj=obj@entry=0x555557259ab0) at kernel/qmetaobject.cpp:378
#2  0x00007ffff5176302 in qobject_cast<QWidgetWindow*>(QObject*)
    (object=0x555557259ab0)
    at ../../include/QtCore/../../src/corelib/kernel/qobject.h:519
#3  QApplicationPrivate::notifyActiveWindowChange(QWindow*)
    (this=<optimized out>, previous=<optimized out>)
    at kernel/qapplication.cpp:2104
#4  0x00007fffecd4a555 in QGuiApplicationPrivate::processActivatedEvent(QWindowSystemInterfacePrivate::ActivatedWindowEvent*) (e=<optimized out>)
    at kernel/qguiapplication.cpp:2518
[...]

Issue #847, opened 2023-12-12
Wayland: Cannot interact with all other application windows when showing the desktop


maehne:
I am not sure whether OpenBoard can do anything about it. Nevertheless, I would like to report that if OpenBoard gets executed on Wayland (using `QT_QPA_PLATFORM=xcb` to use XWayland for correct window placement), the desktop mode works to annotate the screen. However, when selecting the arrow in the tool bar to interact with the other applications, this works only for some windows. Other windows don't receive any input -- apparently because they are behind the transparent window used for rendering the annotations.

I installed OpenBoard 1.7.0 on an up-to-date Endeavour OS (based on Arch Linux, x86_64 processor platform) installation using the [AUR PKGBUILD](https://aur.archlinux.org/packages/openboard).

PR #1165, opened 2024-11-18
fix: avoid some wayland problems

letsfindaway:
As distributions move from X11 to Wayland as default, more and more people are using OpenBoard on Wayland. However there are some known problems, e.g. the following:

- #843, #1041
- #1164
- No screenshots and no screen mirroring in Desktop mode due to Wayland restrictions

The first two of these problems can be mitigated by using the `xcb` backend to force using `xwayland`. This PR sets the corresponding environment variable to enable the xcb` backend:

- use QT_QPA_PLATFORM=xcb when running on wayland
- avoids crash when using the screen configuration list
- avoids invisible window on Qt 6.8

Due to the severity of these problems I will include this PR in all by builds in the future and recommend to merge it as it has no impact for X11 sessions or for other operating systems.

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -96,6 +96,14 @@ int main(int argc, char *argv[])
     if (qEnvironmentVariableIsSet("QT_NO_GLIB"))
         qunsetenv("QT_NO_GLIB");
 
+    // force using xwayland when running on wayland
+#ifdef Q_OS_LINUX
+    if (qgetenv("XDG_SESSION_TYPE") == "wayland")
+    {
+        qputenv("QT_QPA_PLATFORM", "xcb");
+    }
+#endif
+
     Q_INIT_RESOURCE(OpenBoard);
 
     qInstallMessageHandler(ub_message_output);


qqqqqqqqq9:
Hi,
with opensuse package (1.7.2) the Desktop mode is not working.
I only get a black screen.
When I build openboard myself, the desktop mode works.
When I start opensuse with 

QT_QPA_PLATFORM=xcb  /usr/bin/openboard

I get a black background again.

qqqqqqqqq9:
https://github.com/OpenBoard-org/OpenBoard/issues/847

letsfindaway:
I cannot reproduced the black background on my side (Leap 15.6, but also on a tumbleweed distrobox). I only have a black screen when activating screen mirroring, but yeah, a lot of wayland related issues remain. 

My observations:

- **Without** `xcb`, Desktop mode also shows the desktop for me, but the Desktop tools palette on the left side is hidden, as soon as I interact with an application. I have to reveal it by clicking on the OpenBoard taskbar icon - which is a Wayland icon. And then I can draw. But as soon as I use the pointer again to click through, the palette and the drawing vanishes. And when I try to configure my screen setup, OpenBoard crashes as soon as I click on the input field for the list of used screens.
- **With** `xcb` Desktop mode shows the desktop, but I cannot interact with it. I can draw something, but cannot click through to the applications. 
- In both cases, screenshots and screen mirroring only get a black screen. Also podcast recording does not work for desktop mode.

My intention for using the `xcb` backend by default was the crash reported in #843. Let me phrase it this way:

We do not have a good solution for wayland. We can only try to make OpenBoard as usable as possible. And Desktop mode is broken with wayland anyway. So I have chosen what I think is more usable. And a only halfway working desktop mode was the lower price compared to a crash when configuring screens.

Anyway we recommend to use OpenBoard on an X11 session to use all features.

letsfindaway:
Another note: I think that the Desktop mode as we have it in OpenBoard wit X11 can never work on Wayland the same way.

OpenBoard places a transparent, invisible widget on top of the desktop. It can intercept events at will: Depending on the tool they are forwarded to the application below or used for drawing, while OpenBoard still stays in the foreground. This is something the security concept of wayland definitely wants to inhibit, as it would allow any kind of applications stealing user data.

You can already observe what Wayland does when you do not use `xcb`: When clicking on an application it brings this application to the foreground unconditionally. That's the reason why the OpenBoard tool palette and drawings disappear - they are now just behind the focus application. On X11, OpenBoard can still remain in the foreground in this situation.

If OpenBoard is in the foreground, as we have it with `xcb`, then it will not deliver any event to the applications behind. Therefore we cannot interact with them.

So that mixed mode we have on X11, where we can draw n the glass pane in front of the other applications while still delivering events through the "holes" outside of the drawings is not feasible with Wayland.

I started some work on using the Desktop Portal for screenshots and screen mirroring (https://github.com/letsfindaway/OpenBoard/tree/feat-wayland-screenshot) but stopped that again. First we would have to decide: What will be the future of Desktop mode on Wayland? Should we just accept that it will not work as before and completely remove it from OpenBoard? Do we have an idea for a desktop mode which can work with Wayland? I do not have an answer to those questions.

qqqqqqqqq9:
BTW: Do you have multi-monitor mode enabled?
It's greyed out but looks pressed down.

My observations (without xcb)

I can 
  - open konsole (wayland-App)
  - start geoenzo (Xwayland-App, via wine)
  - use geoenzo normally (after disabling fullscreen)
  - use OpenBoard to draw on the geoenzo-window
  - use geoenzo agan.
In my short test, it worked flawlessly.

Screenshot don't work with with openboard, 
however I could use spectacle as a workaround.

> We do not have a good solution for wayland. We can only try to 
> make OpenBoard as usable as possible. And Desktop mode is 
> broken with wayland anyway. 

Is there a way to force OpenBoard to wayland-mode?
QT_QPA_PLATFORM=wayland /usr/bin/openboard didn't work.

> You can already observe what Wayland does when you do not 
> use xcb: When clicking on an application it brings this 
> application to the foreground unconditionally. 

That's not true. For me it works as intended. 
Perhaps your focus-policy is interfering? I use kwin_wayland with Focus-follows-mouse and window-raising disabled.

> First we would have to decide: What will be the future of Desktop 
> mode on Wayland? Should we just accept that it will not work as > before and completely remove it from OpenBoard? 

IMO this is a No-brainer. Being able to draw on other apps so easily is a key feature of OpenBoard.

letsfindaway:
> Perhaps your focus-policy is interfering? I use kwin_wayland with Focus-follows-mouse and window-raising disabled.

Thanks for this hint. Yes, with your settings things look better. But how could we force users to use such a setting? What with other desktop environments, especially Gnome?

Still we have that crash. It now no longer appears directly when clicking on the input field, but when I click on one of the monitor indicators, which now however appear at the wrong location. I have that stack trace with the SEGFAULT:
grafik

qqqqqqqqq9:
> But how could we force users to use such a setting?

Here's what I would do:
1. create a wrapper-script which checks for QT_QPA_PLATFORM. If it's not set to wayland, it starts with QT_QPA_PLATFORM=xcb. That gives choice to users.
2. Explain the wayland/xcb-situation on openboard.ch. State that wayland works better with some setups and encourage to try it.
3. Open a discussion to collect setups which work with wayland and which do not.

You can't force people change their setup but you can give them choice. Also there might be setups in gnome which work well.
It's highly inconvenient to have a setup which doesn't work well, when you in front of 25 12-your-olds. This tends to motivate to change your settings.

As I understand, the crash can be easily avoided by not clicking into the mentioned field? I would add a "Not-TODO"-List.

I can write a draft for the section on openboard.ch.

letsfindaway:
Thanks for all your comments, which gave me new insights into the problem field. I just converted this PR to Draft and am currently in favor not to merge it at all.

> Here's what I would do:
> 
> 1. create a wrapper-script which checks for QT_QPA_PLATFORM. If it's not set to wayland, it starts with QT_QPA_PLATFORM=xcb. That gives choice to users.

Together with your point 2., I would now even completely omit this and let it to the users and the explanation.
 
> 2. Explain the wayland/xcb-situation on openboard.ch. State that wayland works better with some setups and encourage to try it.
> 3. Open a discussion to collect setups which work with wayland and which do not.

I think Step 3 comes before step 2. So yes, feel free to open a discussion about his topic!

> You can't force people change their setup but you can give them choice. Also there might be setups in gnome which work well. It's highly inconvenient to have a setup which doesn't work well, when you in front of 25 12-your-olds. This tends to motivate to change your settings.
> 
> As I understand, the crash can be easily avoided by not clicking into the mentioned field? I would add a "Not-TODO"-List.

I will investigate that further. If we don't find a solution, then I tend to disable the display of those screen indicators as workaround. They are even not correctly positioned on Wayland and therefore currently completely useless.

> I can write a draft for the section on openboard.ch.

First start with the discussion here. I think the result may also go into a Wiki page here. But this is up to the maintainer.

letsfindaway:
Closing this PR as it does not provide an appropriate solution to the Wayland problems.

PR #1168, opened 2024-11-21
Feat: rotate snap indicator and snap rotated items

letsfindaway:
This PR applies the snap-to-grid function correctly to rotated items. It tries to make snapping more foreseeable and toi better fit user expectations for rotated items.

- When moving an arbitrarily rotated item the corners of the item snap. Snap indicators are now correctly shown at the item corners. Until now, a scene bounding rectangle of the item snapped and the snap indicators have been shown on that virtual, invisible rectangle.
- Snapping is now also correctly applied to flipped items.
- Resizing of rotated items can now also snap to the grid if the angle is 0°, 90°, 180° or 270°. For other angles it makes no sense to snap resizing.
- Scaling of an item (i.e. resizing with the bottom-right handle) never snaps.

See also https://github.com/letsfindaway/OpenBoard/issues/185 for a discussion of this feature and the list of tests I have conducted.

### DIFF
--- a/src/domain/UBGraphicsDelegateFrame.cpp
+++ b/src/domain/UBGraphicsDelegateFrame.cpp
@@ -281,7 +281,6 @@ void UBGraphicsDelegateFrame::mousePressEvent(QGraphicsSceneMouseEvent *event)
     mDelegate->startUndoStep();
 
     mStartingPoint = event->scenePos();
-    mStartingBounds = UBGraphicsScene::itemRect(delegated());
 
     initializeTransform();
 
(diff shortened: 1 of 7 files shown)


src/domain/UBGraphicsDelegateFrame.cpp
```diff
@@ -430,13 +451,20 @@ void UBGraphicsDelegateFrame::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
         if(!rotating())
         {
             const auto* ubscene = dynamic_cast<UBGraphicsScene*>(scene());
+            constexpr double epsilon = 0.0001;
 
-            if (ubscene && ubscene->isSnapping())
+            if (ubscene && ubscene->isSnapping() && !resizingBottomRight() && std::fmod(mAngle + epsilon, 90.) <= 2. * epsilon)
```
kaamui, line 456: Why do you disable snapping on bottom right ?

PR #1193, opened 2024-12-29
fix: tool usage with tablet eraser

### LINKED ISSUE #437: Pen pressure not recognized after holding erase button on pen
System: Windows 10, Acer Switch Alpha 12 (with pressure-sensitive pen), version 1.6.1a-1102

On the mentioned version, pressure sensitivity generally works fine (contrary to previous versions, see https://github.com/OpenBoard-org/OpenBoard/issues/302#issuecomment-732837762). Also, as expected, if I hold Button 2 (I think) on the pen, OpenBoard temporarily changes from pen to eraser. 

Steps to reproduce my problem:
1) Select the "Pen" tool and write something on a fresh board using a (pressure sensitive, physical) pen.
2) Lift the pen, press and hold button 2 on the pen, then use the pen to erase something.
3) Lift the pen, then release button 2.
4) Write something on the board.

Expected behavior:
a) The active tool should change back from "eraser" to "pen" when I release button 2.
b) When I write on the board with the pen in step 4, the pen should write normally and with pressure sensitivity.

Actual behavior:
a) The active tool changes from "eraser" to "pen" the moment I touch the screen with the pen.
b) The first stroke after the change does not recognize pressure sensitivity and draws with constant maximum thickness (as if the pen had no pressure sensitivity). Once I lift the pen off the screen and put it back down, everything goes back to normal.

Please let me know if you need further information.

letsfindaway:
This PR tries to fix #437. 

- use correct tool after switching to/from tablet eraser
- re-evaluate `currentTool` after changing it by the tablet eraser

However I need some testers, as I do not have a tablet with an eraser function. I will add this PR as a patch to my experimental builds, so that experimental packages are available later this day.

@bruening-bw : there will be packages for several versions of Debian available. So it would be helpful if you could test whether this fixes the issue for you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,22 +2,22 @@
 ![GitHub Repo forks](https://img.shields.io/github/forks/OpenBoard-org/openboard)
 # OpenBoard
 [![latest release](https://img.shields.io/github/v/release/OpenBoard-org/openboard.svg)]()
-[![Commits since last release](https://img.shields.io/github/commits-since/OpenBoard-org/openboard/v1.7.2/dev)]()
+[![Commits since last release](https://img.shields.io/github/commits-since/OpenBoard-org/openboard/v1.7.3/dev)]()
 [![Github Repo Contributors](https://img.shields.io/github/contributors/OpenBoard-org/openboard.svg)]()
-[![downloads v1.7.2](https://img.shields.io/github/downloads/OpenBoard-org/openboard/v1.7.2/total)]()
+[![downloads v1.7.3](https://img.shields.io/github/downloads/OpenBoard-org/openboard/v1.7.3/total)]()
 [![Github All Releases](https://img.shields.io/github/downloads/OpenBoard-org/OpenBoard/total.svg)]()
 
 OpenBoard is an open source cross-platform interactive white board application designed primarily for use in schools. It was originally forked from Open-Sankoré, which was itself based on Uniboard.
 
 ### Installing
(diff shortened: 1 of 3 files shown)


bruening-bw:
The eraser works as it should. When I use the eraser on the back end of the pen, it works immediately. And if I use the pen again with the tip, it switches back to the drawing tool and is pressure sensitive from the first stroke. This works with the pen and the marker and the straight line tool.

If I have selected the laser pointer and then use the eraser, it does not switch back to the laser pointer afterwards, but to the last drawing tool used before. The same applies to all other tools that are not used for drawing, such as scissors, the selection tool, text, zoom, hand, etc. But that's not a problem. No need to bother with it.

I have also tried the eraser in desktop mode. It works great when erasing on the desktop.

I didn't notice any unwanted side effects during my tests.

letsfindaway:
> If I have selected the laser pointer and then use the eraser, it does not switch back to the laser pointer afterwards, but to the last drawing tool used before. The same applies to all other tools that are not used for drawing, such as scissors, the selection tool, text, zoom, hand, etc. But that's not a problem. No need to bother with it.

This is something which could be changed, too. But I would classify it as a "behavior" and not as a "bug" and would like to have some feedback from @kaamui before changing the behavior.

> I didn't notice any unwanted side effects during my tests.

Thanks for testing. And thanks for bringing up this old topic again - I think it was just forgotten.

So I hope it will be fixed in the next release of OpenBoard.

kaamui:
> This is something which could be changed, too. But I would classify it as a "behavior" and not as a "bug" and would like to have some feedback from @kaamui before changing the behavior.

I'll have to ask too.

PR #1199, opened 2025-01-08
Refactor DnD in Document mode

letsfindaway:
Drag-and-drop in document mode has some issues. Some also affect the user, but others are just implementation issues. This PR addresses both.

### User visible issues

- When dragging a page from the thumbnails to a document, the highlighting of the target row is broken. Both columns (name and date) are highlights independently and the highlight in one column remains when the cursor moves to the other column.
- When dragging a document to another position, the "copy" cursor is shown, even if the document is just moved.

### Internal issues

- The `UBDocumentTreeMimeData` class is weired: it contains a list of indexes as well as a list of URLs, but in the end none of them are ever used.
  - The URLs are not used because they actually are not needed.
  - The indexes are not used, but instead the current selection is used to determine the dragged items. This is an architectural issue. The drop handler should always use the mime data provided.
- Also the mime type is not set in this class.
- The `UBDocumentTreeModel::dropMimeData()` function is never called. Instead all drop handling is already done in `UBDocumentTreeView::dropEvent()`. It is ok to do so: you can either implement dropping in the view or in the model and I'm happy to do it in the view. So that function `dropMimeData()` can be removed.
- This in turn allows to remove `UBPersistenceManager::copyDocumentScene()`, which was only called from that dead code. `dropEvent()` implements scene copying in another way, preferred by me. It loads the scene, determines the related media objects and copies them, saves the scene and copies the related thumbnail.
- This in turn allows to remove the `UBForeignObjectsHandler` used for the other way of copying. This class was parsing the SVG and determines the referenced objects from there. So it implements another SVG parser besides the `UBSvgSubsetAdaptor` and I would like to avoid having two parsers in parallel. It was never proved whether this code actually works because it was never invoked. So I'm happy to remove this class.

### This PR

This PR fixes issues in drag start, drag move and drop:

- When starting a drag in the thumbnails, it allows Copy and Move: For dragging within the thumbnails the Move action is used, while for dragging to the document tree the Copy action is used.
- The `UBDocumentTreeMimeData` class now carries the correct set of indexes (for the first column only) and the correct mime type.
- `dragEnter()` now sets the correct action depending on the mime type.
- `drawMove()` also had some minor issues. We now make sure that the correct cursor is shown indicating the proper drop action (Copy or Move).
- `dropEvent()` uses the mime type and the data contained in the mime data classes.
- Therefore `UBDocumentTreeView::mapIndexesToSource()` is no longer needed, as the indexes in the mime data are already referring to the base model and not the proxy model.
- `UBDocumentTreeModel::setHighLighted()` is fixed to always highlight a complete row. Also updating the UI is now correctly done by emitting a `dataChanged()` signal instead of updating a specific area in `UBDocumentTreeView::updateIndexEnvirons()`.

So finally this PR fixes some user visible issues and at the same time improves and simplifies the code structure of DnD handling in document mode.


letsfindaway:
Converting to Draft as long as #1237 is not merged, as it might need rebasing.

letsfindaway:
I now rebased this branch on top of `feat-toc` as it overlaps to some extend, e.g. removing `UBForeignObjectsHandler`.

Issue #1167, opened 2024-11-20
[Bug] Copying a page to another document moves images to layer 1000000


qqqqqqqqq9:
### Describe the bug

Copying a page to another document moves images to layer 1000000
### To Reproduce

1. Create Doc 1.
2. Add two images to the first page.
3. Create Doc 2.
4. Copy Page 1 of Doc 1 to Doc 2 by drag and drop.

#### Expected behavior
The images should be in Layer 1 and 2, respectively.

#### Actual behavior
The images are in Layer 1e+6. 

Is there a way to move to a layer specific layer?
Doing so by clicking takes a looooong time.

### Context

  - reproducibility : `systematic` 
  - OpenBoard version :
      - bug happens with : `1.7.2`
      - bug doesn't happen with : ``
      - didn't check : `1.6.4` `1.7.0` `1.7.1a-x.y.z`
  - OS : 
      - bug happens with :  `Linux`
      - bug doesn't happen with : 
      - didn't check : `Windows` `macOS`
      
I built from Source on opensuse using
https://aur.archlinux.org/packages/openboard
as template. (qt6)
  
### Additional context

- Linux : openSuSE Tumbleweed

letsfindaway:
Thanks for reporting. I assume this bug is not new with 1.7.2, but here since many years.

You have the same effect when you even drag the page to the same document. However if I check the SVG files, I cannot see those strange z-values. The original page and the copy are identical besides the UUID of the page. And actually: when I close and re-open OpenBoard, then the z-values are ok!

My analysis:

- Looking at the SVG files, the attribute is `ub:z-value="-999999.000000"` for the item which has actually z-value 1 and `ub:z-value="-999998.000000"` for the item which has actually z-value 2. There seems to be a offset of -1e6, for whatever reasons.
- This seems to be intentionally, because I find lines like `UBApplication::showMessage(QString::number(1000000 + newZValue));` when the z-value shall be displayed when pressing the up/down buttons.
- I can also see that there are two places where the z-value is stored in the items:
  - The `zValue` property of the item, and
  - The `UBGraphicsItemData::ItemOwnZValue` data property of the item.
- When a scene is copied by dragging it to a document in document mode, then a clone is created, which is then saved to the target document. This clone is also inserted to the scene cache and is used as long as OpenBoard is running and the cache is not full.
- **When cloning, the items `zValue` properties are copied, but the `ItemOwnZValue` data property is not.**
[...]

PR #1204, opened 2025-01-10
feat: allow symlinks in library

letsfindaway:
This PR addresses a small remark in https://github.com/OpenBoard-org/OpenBoard/issues/1167#issuecomment-2489209363: It is currently not possible to use symlinks within the library folders, e.g. for images, video, audio or interactive content. 

When using symlinks, there is a risk to create cyclic or duplicate structures . This might have been the reason to exclude symlinks when scanning the file system for such features.

This PR handles symlinks properly, still avoiding cacles and duplicates:

- return symlinks in allElementsInDirectory
- use a common file system walker for counting and scanning
- skip symlinks leading to duplicate or recursive scans
- remove obsolete non-threaded functions

See also https://github.com/letsfindaway/OpenBoard/issues/199


kaamui:
I'm still quite afraid of what the potential impact in performances could be in a distant-storage context. 

In local context, did you notice during your tests any change regarding execution time of the scanning ?

Issue #1009, opened 2024-06-07
[Bug] Switching pages moves view by slight amount


Vekhir:
### Describe the bug
Switching between pages shifts the view up (the objects and background look like they are moving down).
It's only by a small amount (a pixel or so per switch), so could be rounding error which accumulates.

### To Reproduce

1. Open OpenBoard
2. Create second page
3. Draw a line on each page or simply turn background on
4. Switch between the pages repeatedly by clicking on their thumbnails

#### Expected behavior
The view doesn't change.

#### Actual behavior
The objects "move down" - since the background moves too, I guess the view is moving up, though I don't know how that looks in the code.

### Context

  - reproducibility : `systematic`
  - OpenBoard version :
      - bug happens with : `1.7.1`, 1665e2e570e5e76c75d9e8a4b7e9a3f0c5071736 (dev)
      - didn't check : `1.6.4` `1.7.0`
  - OS : 
      - bug happens with : `Linux` (own tests) 
      - didn't check :`macOS`  `Windows`
  - Installation source : `own build` via AUR
  
### Additional context
OS: Arch Linux
Kernel: Linux 6.9.3-arch1-1

a) number and resolution of monitors : 2 x 1920x1080, 
b) number and resolution of monitors: 1 x 1920x1080

Vekhir:
Thanks, of course I only tested with Qt6... Qt5 is fine, so I guess that's what you used?

I actually found a simpler reproducer, just clicking on the thumbnail already moves the view/objects/background.

Qt 6.7.1 is affected.

letsfindaway:
Yes, I used Qt 5.15.8 for testing. But I now tested it also with Qt 6.4.2 and did not observe this move.

Does this occur only when you switch pages with the thumbnails or also if you use other methods like toolbar icon or keyboard? This could probably help to identify the code path causing that.

Vekhir:
I've come back to this issue and checked for the exact version where the issue starts. Every version starting with `6.6.2` has this issue (tested 6.6.2, 6.7.0, 6.7.1, 6.7.2). Previous versions are fine (tested 6.6.0, 6.6.1).

[ALA](https://wiki.archlinux.org/title/Arch_Linux_Archive) snapshot 2024-02-15 is fine, ALA snapshot 2024-02-16 exhibits this behaviour. Only difference is Qt 6.6.1 -> 6.6.2.

letsfindaway:
[...]

PR #1220, opened 2025-02-05
fix: small scene shifts when switching pages

letsfindaway:
This PR fixes the small shift of scenes when switching pages initially reported in #1009. Even if the effects of this problem depend on the Qt version use, it is an unclean procedure in the OpenBoard code that causes this.

There are two methods to move the visible region of a `QGraphicsView`:

- We can change the value of the horizontal and vertical scroll bar. These values are integers, so we can only move by pixels.
- Or we can change the transformation matrix of the view. This is the transformation from scene coordinates to view coordinates and everything is floating point here.

OpenBoard uses both, and that is not a good idea:

- When using the Hand tool, only the transform is changed.
- When using the zoom, transform and scrollbar values are changed.
- When using the mouse wheel, only the scrollbar values are changed.
- When switching pages, the viewport center is saved and a combination of `centerOn()` (changing scrollbar values) and `translate()` (changing transformation) is used to restore it.

Even with this PR scrollbar values and transformations are used, but they are clearly separated. A first idea of using transformations only was skipped, because the scroll area of the `QGraphicsView` cannot be guaranteed to never scroll. E.g. the mouse wheel produces an event, which is processed by the scroll area and affects the scrollbar values.

To remember the position and zoom factor of a scene, a `SceneViewState` is used which already provides the necessary information. It was however not consistently used. 

Here is a list of tests I have successfully performed with the updated code:

- [x] Switch pages with prev/next
- [x] Switch pages with first/last
- [x] Switch pages with thumbnails
- [x] Switch pages using Document mode
- [x] Switch zoomed page
- [x] Switch page moved by hand tool
- [x] Switch page scrolled by wheel
- [x] Clear page
- [x] Double-click and long-click with hand
- [x] perform all tests with second display
- [x] Perform all tests with Qt5 and Qt6

Technical details of the changes:

- visible scene area is affected by scene transformation and the view scrollbar values
- clearly separate changes to those two
- save and restore both using the `SceneViewState`
- `UBBoardController`:
  - always use `centerRestore` when re-centering a scene
  - `UBBoardController::centerOn` only affect transformation
  - fix keeping the scene point when zooming
  - save scrollbars and transform in `persistViewPositionOnCurrentScene`
  - add a `restoreViewPositionOnCurrentScene`
  - save and restore view position in `setActiveDocumentScene`
  - remove redundant `saveViewState`
- `UBGraphicsScene`:
  - remove no longer used `lastCenter` and `setLastCenter`
  - remove associated no longer used `SceneViewState` functions

### DIFF
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -550,7 +550,6 @@ void UBBoardController::stylusToolDoubleClicked(int tool)
     else if (tool == UBStylusTool::Hand)
     {
         centerRestore();
-        mActiveScene->setLastCenter(QPointF(0,0));
     }
 }
 
@@ -853,8 +852,7 @@ void UBBoardController::clearScene()
     {
         freezeW3CWidgets(true);
         mActiveScene->clearContent(UBGraphicsScene::clearItemsAndAnnotations);
-        mActiveScene->setLastCenter(QPointF(0,0));
-        mControlView->centerOn(mActiveScene->lastCenter());
+        centerRestore();
         updateActionStates();
     }
 }
@@ -959,7 +957,6 @@ void UBBoardController::zoomRestore()
     centerRestore();
 
     emit zoomChanged(1.0);
-    UBApplication::applicationController->adjustDisplayView();
 
     emit controlViewportChanged();
     mActiveScene->setBackgroundZoomFactor(mControlView->transform().m11());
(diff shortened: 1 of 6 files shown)


kaamui:
Hi @letsfindaway,

could you please rebase this PR to the current state of the `dev` branch please ? 

Thank you

letsfindaway:
> Hi @letsfindaway,
> 
> could you please rebase this PR to the current state of the `dev` branch please ?
> 
> Thank you

Done!

Issue #991, opened 2024-05-27
OpenBoard podcast feature


DylanBruce99:
I'm trying to record some videos on openboard by using the "podcast" function. But the videos are super low frame. I opened the recording in VLC and they are at 3 fps whereas if I record the screen with the MAC's built in feature is almost 60 fps. I'd like to use OpenBoard's because it records automatically the who're writing area and gets rid of all the tools UI. Is there a way to improve the fps? 

Thanks

kaamui:
Hi,

Yes. Add to your OpenBoardUser.config file (located in `$HOME/Library/Application Support/OpenBoard/`) the following entry
```
[Podcast]
FramesPerSecond=?
```
replace `?` by what you want (something your mac can handle). Default value is 10, so maybe there is another issue... do you experience some slow downs when writing with the pen or marker on the board ? on what OS version are you ?

Issue #106, opened 2017-10-02
Cannot set fps higher than 10 on podcast recording videos WMV


rbolrey:
I try to record a podcast, setting frames per seconds to 25, 20, 12 fps, but in all cases the records are 10fps.
I try it in Windows 7 and windows 10, changing the line [Podcast] Frameperseconds=25 in etc/OpenBoard.config
In the same way, can we will use ffmpeg instead windowsmediacodec?

freeheel3r:
Hi,
Thanks for reporting this. It does seem like that setting isn't actually used anywhere, so the frame rate stays at the default 10fps.

Concerning ffmpeg, while it should be technically possible given that we have implemented ffmpeg-based recording on Linux and that the libraries are cross-platform, it would take a considerable amount of work to make everything build & package correctly, so it is not on the roadmap for now.

rbolrey:
Thank you for your great job. OpenBoard is fantastic tool. I use with Clevertouch Pro Plus to record screencast with voice in my University.
While a solution to this issue is founded, I tried to compile Openboard for use in Windows, but always get some errors compiling. (I'm trying it in Ubuntu 16, using -spec win32-g++ follow all the steps, but in thirdparties, the documentations says that use nmake, but in ubuntu don't have any nmake).

I need compile Openboard with two lines modified in /src/core/UBSettings.cpp
    podcastFramesPerSecond = new UBSetting(this, "Podcast", "FramesPerSecond", 10);
    podcastVideoSize = new UBSetting(this, "Podcast", "VideoSize", "Medium");
to:
    podcastFramesPerSecond = new UBSetting(this, "Podcast", "FramesPerSecond", 25);
    podcastVideoSize = new UBSetting(this, "Podcast", "VideoSize", "Full");

Anybody who used to compile, can compiling Openboard to use in Windows for me? 

Thank you very much!

freeheel3r:
Hi,
If you want to build OpenBoard for Windows, you need to do so in Windows -- you can't build a Windows executable from Ubuntu.

And about those lines: changing them will change the default settings, but in the case of the frames per second setting, it won't change anything. The issue here is that the screencast functionality just doesn't take this setting into account. So the fix will be more involved than just changing a setting, unfortunately.

PR #1232, opened 2025-02-19
feat: User-defined podcast fps

Vekhir:
Currently, the podcast recording is set to a fixed 10 fps. While there exists a setting to allow for customisation, it remains unused.

This change takes the user setting into account. Before starting the recording, the frame rate is taken from the settings (as are the other values). The actual fps are determined by the activity on the board, so it represents a maximum fps. 

The default has been bumped up to 100 fps. During testing, I mostly got 40-70 fps in the resulting video. I don't know where the discrepancy comes from, but even setting it to a ridiculous 1000 only achieved like 127 fps peak. For highest quality I'd probably set it to 250 or 300.
One the low side, a value of 25 is at the lower limit of what is perceived as fluid motion. In high activity scenes, it's actually around 12-18, though I regularly hit the 25. If bandwidth or CPU power needs to be saved, the value can be lowered further.
In moments of low activity, all values can go down to 8-10 fps. The horologe produces consistently 20fps.

In conclusion, I believe that this change improves the podcast experience.
Solves #991 #106


kaamui:
Hi @Vekhir,

could I ask you to expose the setting in the preferences ? ~~and to add information in the label (e.g if the setting is actually max 100 fps, to display next to Max, medium and small the corresponding values you calculate) ? Something like "Small (max fps : 25)"~~

~~What do you think ?~~

Edit : expliciting the fps will give the impression the small, medium and maximum aren't about resolution. 

Also, I wonder if it makes sense to attach a FPS limit to the resolution of the podcast. Shouldn't we simply add a section in the dropdown list for the max FPS, independant of the resolution section ?

Vekhir:
> Edit : expliciting the fps will give the impression the small, medium and maximum aren't about resolution.

Yes, the settings are really about the bandwidth/bitrate of the final video. The resolution and framerate are scaled to retain a passable quality per frame when lowering the bitrate.

> could I ask you to expose the setting in the preferences ?

I think the general preferences dialog is quite crowded. Not sure how well that would integrate.
Maybe we could get away with a small dialog box that can be accessed via the recording preferences (like a little edit pen next to "Small", "Medium", and "Full"). There, each preset could be edited independently with bitrate, resolution, and framerate.

I'll try making a mockup. I can use the [intranetPodcastPublishingDialog.ui](https://github.com/OpenBoard-org/OpenBoard/blob/master/resources/forms/intranetPodcastPublishingDialog.ui) as a template and would use the [QSpinBox](https://doc.qt.io/qt-6/qspinbox.html). I guess a dropdown menu could also work, though we'd need to decide which values to show.

kaamui:
> I think the general preferences dialog is quite crowded

Fair point. I like your idea of a dialog directly from the record preferences. I don't how much effort it would take, but maybe this cog button could directly open a dialog box with the audio and video settings, and no longer be a drop down menu where we are limited in terms of space and ux. Could you give it a try ?

Vekhir:
I'll see what I can do. Will create a separate PR for that (focusing on the reworked dialog), then adjust this PR to add the new settings.

Vekhir:
Just to let you know, I currently don't have time to work on this. I'd like to see it come to fruition, but I can't commit the necessary time and energy.
Once I resume working on this issue, I'll write here. In the meantime, everyone feel free to work on this if you want.

Vekhir:
I've created a new PR #1472 that supersedes this one. It includes a completely reworked UI for setting the preferences and incorporates the fixes from this PR.

Vekhir:
I've decided to split the other PR to only include the new preferences dialog. The actual fixes are again in this PR, keeping the two changes separate. This PR builds on top of #1472.

Vekhir:
I'm reopening this PR

PR #1242, opened 2025-03-12
Refactor enable UI

letsfindaway:
This PR consists of two commits, addressing architectural issues.

- The first one moves the information whether the thumbnail UI (duplicate and delete buttons) should be visible or not to the respective `UBThumbnailArranger`. This keeps dependencies to specific views out of the `UBThumbnail` class.
- The second one moves `UBThumbnailTextItem` out of the `UBDocumentThumbnailsView` header and creates separate files for this completely independent class.


kaamui:
Thank you

Issue #1278, opened 2025-06-04
[Bug] *.WEBP image file is not supported


sebojolais:
### Describe the bug
If you copy paste the URL of a WEBP image in an empty page of the Board mode, nothing is displayed.

### Background

I am working on an application where I converted a list of PNG images to the WEBP format in order to reduce the file sizes. And this app, I need to add them on the board via the API `window.sankore.addObject()`, but it does not work.

### To Reproduce

1. Open OB
2. Go to the Board mode, in a blank page
3. Copy this link : [https://en.wikipedia.org/wiki/File:Wikipedia-logo-v2-webp.webp](https://en.wikipedia.org/wiki/File:Wikipedia-logo-v2-webp.webp)
4. Paste it in the blank page of OB
5. You can see nothing, no image added
6. I expect to see the image

It is strange because if you import the same file via the Document mode, it works.

### Solution

In the method: `UBFileSystemUtils::mimeTypeFromFileName()` , we have to add this line:
`    if (ext == "webp") return "image/webp";
` 

And in the `UBFileSystemUtils::mimeTypeFromString()` method , if we add:
`        || typeString == "image/webp"
`

Then this issue is fixed.

### Context
  - reproducibility : `systematic`
  - OpenBoard version :
      - bug happens with : `1.7.3`
      - didn't check : `1.6.4` `1.7.0`
  - OS : 
      - bug happens with : `Linux`
      - didn't check : `Windows` `macOS` 
  - Installation source : `own build`
  - If you built OpenBoard from source, please provide the branch you're working on : `dev`

sebojolais:
The PR #1279 is merged. I close this issue.

PR #1279, opened 2025-06-04
Webp image support

sebojolais:
Fix this issue: #1278


sebojolais:
This PR leads me to ask a question :
- Are the the method  https://github.com/OpenBoard-org/OpenBoard/blob/9a8abac5c087256b5bfefa6255fac45481bd5e7d/src/frameworks/UBFileSystemUtils.cpp#L415-L489 
- and the method https://github.com/OpenBoard-org/OpenBoard/blob/9a8abac5c087256b5bfefa6255fac45481bd5e7d/src/frameworks/UBFileSystemUtils.cpp#L492-L562

are really usefull ? 
Because since Qt 5, there are these methods that do the job for us in an exhaustive way:
- [`QMimeDatabase::mimeTypeForFile()` can return the `mimetype` from filename](https://doc.qt.io/archives/qt-5.15/qmimedatabase.html#mimeTypeForFile-1)
- And [`QMimeType::suffixes()` return the extension of a `mimetype`](https://doc.qt.io/archives/qt-5.15/qmimetype.html#suffixes-prop)

kaamui:
Hi @sebojolais,

Maybe these functions could be simplified or extend Qt ones, but removed I'm not sure because of our own extensions and filetypes

sebojolais:
> Hi @sebojolais,
> 
> Maybe these functions could be simplified or extend Qt ones, but removed I'm not sure because of our own extensions and filetypes

Yes it is true, we have to keep them.

But there are some benefits to change the implementation by using QT API for common mime-types and keeping the parts for our own format:
- Any issue like mine with a file type not supported will be fix if it is supported by Qt
- No more code to support new standard, it will be done by Qt
- We have to maintain our code just to support our own format, no more stuff for standard.

More details about the formats that OB will support after the modifications. It seems that since Qt 6, the mimetypes detection is provided by the freedesktop.org shared-mime-info project. https://doc.qt.io/qt-6/qmimedatabase.html. So the numbers of file format supported by the function will be wide.

Are you agree to modify these functions in this sense ? If yes, I can provide a PR.

Issue #1270, opened 2025-05-20
[Announcement] Preview of 1.7.4 for Linux


letsfindaway:
The next version of OpenBoard, 1.7.4, is feature complete and nearly ready for the first release candidate. To give users an early preview and allow testers to check out the next version, I have created packages for openSUSE, Debian, Ubuntu, and Arch.

Check out [my experimental builds](https://github.com/letsfindaway/OpenBoard/issues/161)! Any feedback or bug reports are welcome. A Windows preview version will probably arrive around end of June.

Version 1.7.4 is primarily a bug fix release. The only new feature is an enhancement to the snap-to-grid feature, which now works with rotated objects as well.

The most significant internal change is the background loading of thumbnails, which greatly improves performance for large documents, particularly those stored on a network. Thumbnails are now cached, so selecting a document a second time is instantaneous.

Changes in 1.7.4:

- [Performance] Load thumbnails in background. Avoid any waiting time when selecting large documents in Document mode.
- [Feat] Snap-to-grid with rotated objects.
- [Fix] Small scene shifts when switching pages (#1009, #1176, #1218).
- [Fix] Duplicate page on PDF import (#1156).
- [Fix] Resize and rotate horizontal or vertical lines (#1105).
- [Fix] Select previous tool after using pen eraser on tablet (#437).
- [Fix] Drop PDF URL to Board (#1209).

bruening-bw:
I tried the Debian 12 package. It worked fine. I confirm the bugs listed before are solved. I did not find any unexpected behavior. I will test it further and report back if I encounter any trouble.

Tank You so much for this contribution to a better World!!!

fadikdawoud:
## First of all, thank you for the amazing work on version 1.7.4

However, I’ve encountered two issues on `Windows 11`:

#### Object Resizing with Pen tablet Input
In previous versions, it was possible to resize selected objects using a pen on a drawing tablet (as well as the mouse). In version 1.7.4, resizing still works with the mouse, but no longer responds to pen input. It just moves it all over

#### Crashing When Navigating Quickly
Sometimes (not always), the application crashes when switching between older documents or flipping pages fast or even switching between desktop and board.

I’d really appreciate it if you could me fix it by telling me which .cpp files to look into and fix.

bruening-bw:
When I read about the problems, I checked the Linux version to make sure I did not overlook anything. I was relieved to see that I could not reproduce the errors on Linux. Everything seems to be okay.
[...]

PR #1288, opened 2025-06-19
fix: detect delegate handle with tablet pen

letsfindaway:
This PR tries to fix https://github.com/letsfindaway/OpenBoard/issues/221 as reported here https://github.com/OpenBoard-org/OpenBoard/issues/1270#issuecomment-2954111653.

- do not use `isUnderMouse()` as this does not work with tablet pen (in some cases)
- instead determine delegate handle from position using `contains()`

We still do not use the function calls like `rotateButtonBounds().contains(pos)`, as these do not guarantee a complete alignment between the cursor shape and the actual action performed when the mouse button is pressed.

This PR is tested with a Wacom One pen display. However I do not have and cannot reproduce the problem described by @fadikdawoud with this device even with the current `dev` branch. So I would like to ask @fadikdawoud whether my PR works for you.

### DIFF
--- a/src/domain/UBGraphicsDelegateFrame.cpp
+++ b/src/domain/UBGraphicsDelegateFrame.cpp
@@ -315,7 +315,7 @@ void UBGraphicsDelegateFrame::mousePressEvent(QGraphicsSceneMouseEvent *event)
 
     mOriginalSize = delegated()->boundingRect().size();
 
-    mCurrentTool = toolFromPos(event->pos());
+    mCurrentTool = toolFromPos(event->scenePos());
 
     if (mCurrentTool == Rotate)
     {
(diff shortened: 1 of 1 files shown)


fadikdawoud:
Hey @letsfindaway I having been trying it out for a couple of days now to make sure everything is working fine and It's safe to say that it works wonderfully fine, thank you so much

I really am thankful for the work you do and the effort you put in, you're a true inspiration!!!!

letsfindaway:
> Hey @letsfindaway I having been trying it out for a couple of days now to make sure everything is working fine and It's safe to say that it works wonderfully fine, thank you so much
 
Thanks for testing! Then I take this as confirmation of this PR and think it should be part of 1.7.4.

PR #1290, opened 2025-06-24
ADD new shapes directory: cuisenaire_rods

Gvf63:
A directory with shapes used in primary school


kaamui:
Hi, thank you for this proposal. I don't know how many times it will take, but I plan to change the way applications, images, and even shapes can be found. 

The idea is not to increase the amount of data added to the initial package, but to create an easy way to find new content, like yours, when in the application, using a search bar in the shapes folder for example, or by adding new web search engines, etc..

There is also an idea to have degree-oriented or subject-oriented packages/plugins, that could be added to the application (like a math plugin, adding new shapes, tools, etc to the application when installed). These plugins could be handled in a seperate reporistory of OpenBoard, dedicated to widgets and other educational content

So, I must decline your pull request, but I hope to be able to convince the OpenBoard decision board that we should "shape" the future of OpenBoard in this way, where your work could be integrated

Issue #679, opened 2022-10-06
Request: Support screenshot in Desktop mode under Wayland


letsfindaway:
Under Linux/Wayland, the screenshot feature in Desktop mode just creates black rectangles. See #665. The reason for this behavior is, that Wayland does not allow to grab arbitrary region of the screen without explicit user consent. This also affects mirroring of the desktop in multi-monitor setups and podcast recording.

Instead of using `QScreen::grabWindow()`, we have to use a DBus service to take a snapshot.

### Flatpak

Flatpak is a containerized application format. To take screenshots from the sandbox, the `org.freedesktop.portal.Screenshot` interface is specified. Some pointers:

* https://flatpak.github.io/xdg-desktop-portal/ (see especially org.freedesktop.portal.ScreenCast and org.freedesktop.portal.Screenshot)
* https://doc.qt.io/qt-5/qtdbus-index.html
* https://github.com/flatpak/xdg-desktop-portal
* https://community.kde.org/Guidelines_and_HOWTOs/Flatpak#For_developers (with the address of a matrix salon to ask for help)

### Others

The portal service described for Flatpak seems to be the basic mechanism, which should also work on Gnome, KDE and others.

### Discussions

Here some pointers to discussions, even if they don't target C++ or Qt:

- https://stackoverflow.com/questions/72216896/how-can-i-get-a-screenshot-on-wayland-with-pure-python
- https://adangel.org/2022/02/06/java-screenshot/
- https://stackoverflow.com/questions/56368170/using-freedesktop-portal-to-take-screenshots-with-python

letsfindaway:
## First experiences

I made some initial tests with the desktop portal with the following findings:

- The screenshot initiated via the desktop portal always covers all connected monitors. There is no option to specify a screen or an area of the screen. You have to cut the relevant area from this overall screenshot.
- The screenshot is saved as PNG file to the local file system. There is no option to specify the folder. On my system (KDE, X11) they are in the `~/Images` folder, even if the documentation says:
> the returned URI will point into the document portal fuse filesystem in /run/user/$UID/doc/

- The desktop portal informs the application about the location of this file. The application can then read this file and probably delete it when done.

[...]

PR #1299, opened 2025-07-28
Feat wayland

letsfindaway:
This PR makes OpenBoard fit for Wayland, which is a significant development. It resolves the currently known issues and limitations and provides corresponding solutions.

The PR is still a draft because it needs more testing and fine-tuning. However, we will encourage testers to provide feedback on its behavior on different platforms and desktop environments.

Each commit addresses a specific issue:

## feat: screenshot with wayland using desktop portal

Wayland does not allow applications to access the content of an arbitrary screen. This affects screenshots in Desktop mode.

This commit introduces the new class `UBDesktopPortal` which contains functions for taking screenshots and starting screen casts. The platform-specific function `grabScreen()` in `UBPlatformUtils` uses the screenshot function on Linux if the application is running in a Wayland session.

Since taking a screenshot with the portal is an asynchronous function, the result is delivered via a callback. In turn, this requires minor modifications wherever this function is invoked.

Taking a screenshot with the desktop portal may prompt a dialogue box asking the user for permission. It may also be accompanied by acoustic feedback. Both can be experienced on Gnome but not on KDE.

## feat: use screencast portal for desktop mode mirroring

The same issue also affects screen mirroring in Desktop mode. Here we use the ScreenCast function of the desktop portal to access a video stream of the screen content.

When starting the screen cast, a dialog box prompts the user to select the screen to share. The user must then select the correct screen.

In terms of implementation, a new class, `UBPipewireSink`, is introduced to receive the video stream and extract the individual images, which are then rendered on the mirror screen. This class also adds a new dependency on the `pipewire-0.3` library.

## chore: update build files

This commit updates the project build files and adds more `#ifdef` statements to limit the changes to Linux hosts.

## feat: podcast in desktop mode

Just as for screen mirroring, we also have to use the screen cast for podcast recording in Desktop mode. 

This commit adds the necessary changes to `UBPodcastController` to start and stop the screen cast.

Note that currently the screen selection dialog pops up when the user first switches to Desktop mode during a recording.

## fix: screen labels on wayland

This commit fixes a bug that caused OpenBoard to crash when the list of screens was activated in the Screen Configuration dialog. It also improves the positioning of the labels.

The crash only occurs when the screen labels are created by the Preferences dialog. Now, we create the labels in the UBDisplayManager and only show and hide them from the dialog.

Wayland only allows screens to be assigned to full-screen windows. In order to properly position the screen labels, it is necessary to first make them visible in full-screen mode and then adjust them to their actual size.

## feat: keep desktop mode palette always on top with KDE

The last commit only applies to KDE. In this case, the desktop palette does not stay on top when another application is activated and overlaps it. I did not observe this problem on Gnome.

For KDE, we add a KWin script that sets the desktop window's property, `keepAbove`, to `true` whenever we switch to Desktop mode. The script is started in `UBPlatformUtils::keepOnTop()` by D-Bus calls if a KDE environment is detected.


letsfindaway:
Now removing the "Draft" status as I don't have any open points for this PR on my list.

kaamui:
Hi @letsfindaway,

thank you for the time and effort you made on this PR. The OS team is considering using Gnome Wayland next year, so all the issues you resolved with screenshot support and podcast is a huge step forward Wayland compatibility. 

I made some quick tests on gnome wayland (Ubuntu 24.04), Qt 6.7.2. I encountered two issues : 

- immediate crash when clicking on the Web Mode : "The Wayland connection experienced a fatal error: Erreur de protocole"
- Desktop Mode not always on top but I guess it's because you resolved it for KDE only

Would it be possible for you to try to solve these issues on Gnome ? I suppose the Web Mode is not crashing or you would have noticed it. 

Do I need to place some file in my build subfolders to make things work ? Again, not a lot of time to read everything so sorry if you explained this somewhere

letsfindaway:
> Hi @letsfindaway,
> 
> thank you for the time and effort you made on this PR. The OS team is considering using Gnome Wayland next year, so all the issues you resolved with screenshot support and podcast is a huge step forward Wayland compatibility.

Indeed I started about 3 years ago with first investigations (#679) and gave up again. In the meantime I learned a lot and also Wayland and Qt have developed, so that a solution seems now feasible.
 
> I made some quick tests on gnome wayland (Ubuntu 24.04), Qt 6.7.2. I encountered two issues :
> 
> * immediate crash when clicking on the Web Mode : "The Wayland connection experienced a fatal error: Erreur de protocole"

My Gnome test environment is a virtual machine with Debian 12. Of course only single monitor, so I cannot test screen mirroring. Here I cannot reproduce the crash. I searched on the Internet and several people reported such stuff. One idea was to set `QT_QPA_PLATFORM=wayland` before starting the application. But in the end all agree that it is some problem between Qt and mutter.

> * Desktop Mode not always on top but I guess it's because you resolved it for KDE only

My problem is more the other way: I cannot interact with applications using the mouse in Desktop mode. I cannot select text, cannot move windows etc. So I also was not able to rise a window above OpenBoard.

> Would it be possible for you to try to solve these issues on Gnome ? I suppose the Web Mode is not crashing or you would have noticed it.

I actually never worked with Gnome and - let me phrase it this way - if I could avoid it, I would like to avoid it. It was hard enough on KDE, but with Gnome I would have to learn another desktop environment from scratch. 

> Do I need to place some file in my build subfolders to make things work ? Again, not a lot of time to read everything so sorry if you explained this somewhere

No, there are no additional files needed. For KDE there is this additional file for the KDE script. With CMake this is already placed in the correct folder when making an `cmake --install`. For Gnome however it does not hurt if this file is missing.

kaamui:
> My problem is more the other way: I cannot interact with applications using the mouse in Desktop mode. I cannot select text, cannot move windows etc. So I also was not able to rise a window above OpenBoard.

If I immediately click on something on the desktop, it works, and OpenBoard disappears. But if I wrote something on the transparent scene, I cannot click through it anymore. 

> I actually never worked with Gnome and - let me phrase it this way - if I could avoid it, I would like to avoid it. It was hard enough on KDE, but with Gnome I would have to learn another desktop environment from scratch.

I understand. I'll discuss about it internally to see what are our options.

letsfindaway:
> > My problem is more the other way: I cannot interact with applications using the mouse in Desktop mode. I cannot select text, cannot move windows etc. So I also was not able to rise a window above OpenBoard.
> 
> If I immediately click on something on the desktop, it works, and OpenBoard disappears. But if I wrote something on the transparent scene, I cannot click through it anymore.

Just tested: No, not even then I can click through to applications. But Ubuntu 24.04 is based on Debian 13. Perhaps this already makes a difference. I was using Debian 12 for the test machine because this is what you mention as "officially supported". Are you now moving to Debian 13 or even back to Ubuntu?
 
> > I actually never worked with Gnome and - let me phrase it this way - if I could avoid it, I would like to avoid it. It was hard enough on KDE, but with Gnome I would have to learn another desktop environment from scratch.
> 
> I understand. I'll discuss about it internally to see what are our options.

Yes, this would need somebody with more experience with Gnome.

jbruechert:
> For KDE, we add a KWin script that sets the desktop window's property, keepAbove, to true whenever we switch to Desktop mode. The script is started in UBPlatformUtils::keepOnTop() by D-Bus calls if a KDE environment is detected.

Hi, have you already talked to Plasma / KWin developers about this by any chance?
This seems like something that should definitely be a wayland protocol, but I'm not knowledgeable enough on this to point you to the right API.
I would not really expect applications to run KWin scripts, and I think adding a flatpak permission to do that would possibly not pass review.

**EDIT**: I think layer-shell might be the correct solution to this. There are already qt-bindings for it: https://github.com/KDE/layer-shell-qt

letsfindaway:
> > For KDE, we add a KWin script that sets the desktop window's property, keepAbove, to true whenever we switch to Desktop mode. The script is started in UBPlatformUtils::keepOnTop() by D-Bus calls if a KDE environment is detected.
> 
> Hi, have you already talked to Plasma / KWin developers about this by any chance? This seems like something that should definitely be a wayland protocol, but I'm not knowledgeable enough on this to point you to the right API. I would not really expect applications to run KWin scripts, and I think adding a flatpak permission to do that would possibly not pass review.

No, I have not. And for Geneva schools we also have to have GNOME on board.

When I read posts like this https://discourse.gnome.org/t/any-way-to-set-window-always-on-top-programmatically/31579/2 from the GNOME team then I doubt that this will ever happen.
 
> **EDIT**: I think layer-shell might be the correct solution to this. There are already qt-bindings for it: https://github.com/KDE/layer-shell-qt

I experimented with this but it is only usable for applications which completely are on top, like an onscreen keyboard. It cannot be mixed with other "normal" windows. At least I could not get it working this way using https://github.com/KDE/layer-shell-qt.

And even if this protocol is implemented by many compositors, Mutter is not amongst them. https://wayland.app/protocols/wlr-layer-shell-unstable-v1#compositor-support. I think GNOME does not like the idea.

jbruechert:
> I experimented with this but it is only usable for applications which completely are on top, like an onscreen keyboard. It cannot be mixed with other "normal" windows. At least I could not get it working this way using https://github.com/KDE/layer-shell-qt.

Unfortunately I lack experience with this, but the KDE screenshot tool (Spectacle) is also using layer shell and has normal windows, so it must be possible I think.
 
> And even if this protocol is implemented by many compositors, Mutter is not amongst them. https://wayland.app/protocols/wlr-layer-shell-unstable-v1#compositor-support. I think GNOME does not like the idea.

This is not different from the current KWin script though, right?

letsfindaway:
> > I experimented with this but it is only usable for applications which completely are on top, like an onscreen keyboard. It cannot be mixed with other "normal" windows. At least I could not get it working this way using https://github.com/KDE/layer-shell-qt.
> 
> Unfortunately I lack experience with this, but the KDE screenshot tool (Spectacle) is also using layer shell and has normal windows, so it must be possible I think.

I looked into the Spectacle source code. LayerShell is used here (and only here): https://github.com/KDE/spectacle/blob/ca0a8a207848d048de7e066bc15cbb58bd1da689/src/SpectacleCore.cpp#L134-L150

You see that the native resources of the window is first released using `destroy()` before it is attached to the Layer Shell. I assume, even Spectacle uses it either-or. And I still assume that only one shell protocol can be active for an application at one time.

> > And even if this protocol is implemented by many compositors, Mutter is not amongst them. https://wayland.app/protocols/wlr-layer-shell-unstable-v1#compositor-support. I think GNOME does not like the idea.
> 
> This is not different from the current KWin script though, right?

Yes, sigh. I have no idea whether GNOME extensions could do something similar. But from following some discussions about this topic I'm quite sure that GNOME will never ever support a Wayland protocol to control window stacking order.

**Edit:** I must admit that I don't fully understand the Spectacle code. What window is it exactly, which is converted to use LayerShell? In which situation? Are there other windows from Spectacle visible at the same time?

jbruechert:
> Edit: I must admit that I don't fully understand the Spectacle code. What window is it exactly, which is converted to use LayerShell? In which situation? Are there other windows from Spectacle visible at the same time?

Fron my understanding it is the capture selection rectangle that stays active during video recording. I don't think any other windows are active at the same time. The code suggests it is per-window though.

letsfindaway:
> > Edit: I must admit that I don't fully understand the Spectacle code. What window is it exactly, which is converted to use LayerShell? In which situation? Are there other windows from Spectacle visible at the same time?
> 
> Fron my understanding it is the capture selection rectangle that stays active during video recording. I don't think any other windows are active at the same time. The code suggests it is per-window though.

Thank you! I think I will look more deeply into it next year. A solution without KWin script would be fine!

However it will most probably not the solution for GNOME. If you have time, read https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/1141 (see also the last comment on using an extension).

jbruechert:
> However it will most probably not the solution for GNOME. If you have time, read https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/1141 (see also the last comment on using an extension).  

I had a quick look. To me this sounds like the GNOME developers so far did not see any need for this and think that it's only relevant for third-party panels and such. Not wanting to support that is understandable imo.
In this case, it might be worth asking them how they would like such a feature to be implemented. I can also ask my GNOME contacts which might lead to better results.

I know from experience that having people demand a specific technical solution in the issue tracker without describing what they actually want to achieve can be a bit unproductive and gives little incentive to actually implement it.

kaamui:
Hello @jbruechert

Internally, the team in charge of our Linux environments has decided to migrate to Gnome+Wayland in the coming years. This is obviously a concern for OpenBoard, as we are not certain we will be able to get all the features working as intended.

If you could, in any way, help us contact the GNOME team and share information about any issues encountered in this project, it would be very helpful. If there are any insurmountable obstacles to the functionality of OpenBoard's desktop mode, or other important features, it is also important that we discover them as soon as possible so that I can inform our team.

We are actively seeking external service providers to assist us with the support of Gnome+Wayland. Please feel free to contact me directly if you know anyone with the necessary skills who might be interested, or if you just know a good contact to start the conversation with.

jbruechert:
> If you could, in any way, help us contact the GNOME team and share information about any issues encountered in this project, it would be very helpful. If there are any insurmountable obstacles to the functionality of OpenBoard's desktop mode, or other important features, it is also important that we discover them as soon as possible so that I can inform our team.

The one GNOME person I contacted does not directly work on the compositor I think, but considered it very unlikely that this use case would be considered relevant in GNOME unfortunately.

He unfortunately thinks it is impossible that layer-shell support will be added..

kaamui:
Can you resolve the conflicts so I can merge it ? Thank you !

letsfindaway:
> Can you resolve the conflicts so I can merge it ? Thank you !

done

PR #1300, opened 2025-08-03
fix: terminate podcast on program exit

letsfindaway:
- When OpenBoard is closed while a podcast is running the application is hanging waiting for a thread to finish.
- Cleanly terminate the podcast recording before closing the application.

Fixes https://github.com/letsfindaway/OpenBoard/issues/223

**Note:** The `lastWindowClosed` signal is not emitted from the application when terminating OpenBoard. We therefore now connect to `UBMainWindow::closeEvent_Signal`. This also affects other locations, but there we always have another way in place to detect termination, so I did not touch them.


letsfindaway:
I now added a timer to avoid endless waiting.

Also note that a similar construct using `processEvents()` is also in the destructor of the `UBPersistenceManager` and seems to work for several years without problem.

@kaamui : If you think it would be useful I can also move the target branch to `1.7.4-dev`-

PR #1335, opened 2025-09-27
Fix input filter in greek translation

blu256:
The wildcard (*.) was missing in the Greek translation, resulting in the first extension of the filter list not being correctly recognized as such.


letsfindaway:
I just checked all the other translations: a similar problem exists in the Danish translation. Here just the dot is missing. Could you probably even fix that? You don't need to speak Danish to do it ;)
https://github.com/OpenBoard-org/OpenBoard/blob/9a884af1cb63afe5c4f9bb4c31dc567d8b8bb257/resources/i18n/OpenBoard_da.ts#L1916-L1917

All others seem to be correct or are not translated at all.

blu256:
Fixed, sorry for the delay.

kaamui:
Hi,

Thank you for your contribution. Added to the `dev` branch

PR #1337, opened 2025-09-30
fix: adjust replace dialog size

letsfindaway:
- `UBDocumentReplaceDialog`  dialog way too large starting with Qt 6
- resize dialog to minimum fitting size

### DIFF
--- a/src/document/UBDocumentController.cpp
+++ b/src/document/UBDocumentController.cpp
@@ -151,6 +151,9 @@ UBDocumentReplaceDialog::UBDocumentReplaceDialog(const QString &pIncommingName, 
     mainLayout->addLayout(labelLayout);
     mainLayout->addLayout(buttonLayout);
 
+    // resize dialog to minimum fitting size, necessary since Qt 6
+    resize({});
+
     acceptButton->setEnabled(false);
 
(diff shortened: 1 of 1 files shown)


kaamui:
Cannot reproduce the issue with Qt 6.9.1 on Windows, nor with Qt 6.7.2 on Ubuntu (tested with dev+1311.patch) 

I just have the replacement dialog misplaced now with Qt 6.7.2

Do you reproduce it on your side ?

letsfindaway:
> I just have the replacement dialog misplaced now with Qt 6.7.2

Where is it? Where should it be? Wayland or X11 or also on Windows?

> Do you reproduce it on your side ?

Yes. It looks as follows here:
grafik
with Qt 6.6.3. So it might be fixed with a later Qt version.

**Edit:** I went through the release notes of 6.7.0, 6.7.1 and 6.7.2 but could not find anything related to this.

kaamui:
Closing because probably not reproducible with newer versions

letsfindaway:
> Closing because probably not reproducible with newer versions

I still have this issue with Qt 6.9.1 here (`dev` branch):
grafik
grafik

Issue #1349, opened 2025-10-20
[Bug] typo in english tr


sebojolais:
### Describe the bug

In `Preferences` dialog, the tab `Licences` should be called `Licenses`.

<img width="267" height="98" alt="Image" src="https://github.com/user-attachments/assets/2c496712-1fc4-4282-b3cf-cf648cf3e5a3" />

### To Reproduce

1. Open OB
2. Click on 'OB' button
3. Select `Preferences`
4. read the tab texts

### Context
  - reproducibility : `systematic`
  - OpenBoard version :
      - bug happens with : `dev` branch
  - OS : 
      - bug happens with : `Linux`
      - didn't check : `Windows` `macOS` 
  - If you built OpenBoard from source, please provide the branch you're working on : `dev`

sebojolais:
It seems that :
- Licences is for UK
- Licenses is for US

Because I do not know what is the source language of OB, I do not know if this issue is really a typo that must be fixed...

letsfindaway:
> It seems that :
> 
> * Licences is for UK
> * Licenses is for US
> 
> Because I do not know what is the source language of OB, I do not know if this issue is really a typo that must be fixed...

The source language of OpenBoard us US English. It is e.g. using "Color" and not "Colour". There are translation files for `en_US` and `en`, but both of them are mainly "unfinished".

sebojolais:
Hi @letsfindaway,
Thank you you for the information.
So what do you think ? If the source language should be en US, must we fix the sources of OB to `Licenses` even if it will add an entry for every translations ? Or trying to fix it in the .ts files ?

PR #1350, opened 2025-10-20
Fixes a typo - Licences should be Licenses in Preferences dialog.

sebojolais:
Fixes the issue #1349.

### DIFF
--- a/resources/i18n/OpenBoard_en.ts
+++ b/resources/i18n/OpenBoard_en.ts
@@ -3791,7 +3791,7 @@ p, li { white-space: pre-wrap; }
     <message>
         <location filename="../forms/preferences.ui" line="1103"/>
         <source>Licences</source>
-        <translation type="unfinished"></translation>
+        <translation>Licenses</translation>
     </message>
     <message>
         <location filename="../forms/preferences.ui" line="3104"/>


Vekhir:
Please don't change the name in the source file - It's the reference for the translations which would all have to be adapted as well.

There are both `OpenBoard_en.ts` and `OpenBoard_en_UK.ts` translation files. I'm not sure whether the "en" version is supposed to be en_US or just a default. In any case, as the difference is US vs UK spelling, you might want to add an explicit `OpenBoard_en_US.ts` translation file for your spelling.

sebojolais:
Hi Vekhir,
The both files `OpenBoard_en.ts` and `OpenBoard_en_UK.ts` are identical.
It does not help us to handle this issue...

sebojolais:
And I agree, this topic should not induce a new entry for every translation files.
So it will be better to fix it in a .ts file.

Vekhir:
As there are bound to be other differences, it might be most useful to copy the `OpenBoard_en.ts` to create an `OpenBoard_en_US.ts` translation.
I'd rather have the default be the British spelling as that is the most commonly taught dialect of English in many parts of Europe, and - I assume - in Switzerland where this software has its origins.

sebojolais:
I changed the implementation.
Currently, the `string` is changed only in `OpenBoard_en.ts` and not in `OpenBoard_en_UK.ts`.
But it is the first difference introduced between the files, so I am not sure that is best way.

sebojolais:
Currently, this PR introduces minimal changes.
I am waiting @kaamui  feedback before changing the name of file.
Even If I agree with you @Vekhir

Issue #1346, opened 2025-10-17
[Building/Packaging] Qt 6.10 compatibility


Vekhir:
**Describe the problem**
With Qt 6.10 being released, there are new warnings showing up during the build. This time, it's 1 warning occuring 6 times:
```
/usr/include/qt6/QtCore/qfile.h:291:32: note: declared here
  291 |     QFILE_MAYBE_NODISCARD bool open(OpenMode flags) override;
      |                                ^~~~
/build/openboard-git/src/OpenBoard/src/board/UBFeaturesController.cpp: In member function ‘void UBFeaturesController::saveFavoriteList()’:
/build/openboard-git/src/OpenBoard/src/board/UBFeaturesController.cpp:586:14: warning: ignoring return value of ‘virtual bool QFile::open(QIODeviceBase::OpenMode)’, declared with attribute ‘nodiscard’ [-Wunused-result]
```

Ignoring the result can lead to weird and inconsistent behaviour as well as difficult to debug issues when the file cannot be opened as expected. This can be the case if the file is intermittently deleted, moved, inaccessible or has the wrong permissions.
At best, the program crashes immediately, at worst, there is silent corruption that is only noticed far later.

**Suggestions, solutions**
Ideally, a user-facing notice (e.g. a popup window) informs the user that the file couldn't be accessed and allows them to retry.
At the very least, a message should be logged in case of failure and the operation aborted.

I can do the latter, but might need guidance for the UI. Which UI design would you prefer?

-- Vekhir

**Versions**
Provide version information for all components and libraries.
- OS: Arch Linux
- Kernel: Linux 6.17.2-arch1-1
- OpenBoard version: `dev`, but likely affects all versions
- Qt version: 6.10.0

**Appendix**
[openboard-git-1.7.4a.070511.r24.g1bcfaea4-1-x86_64-build.log](https://github.com/user-attachments/files/22979744/openboard-git-1.7.4a.070511.r24.g1bcfaea4-1-x86_64-build.log)

letsfindaway:
Thanks for always checking the latest versions! Fixing such things improves code quality and makes OpenBoard more future-proof.

I agree that the return value of `open` should always be checked. However a user notification is mostly useful if the user can do anything about it. Let's just look at the two locations in the `UBFeaturesController`. Here the warning occurs when the favorites list is read and when it is written.

[...]

PR #1362, opened 2025-10-29
Check return value of open()

Vekhir:
The return value of open() should always be checked and handled to avoid difficult bugs. For the most part, they are, but in some cases this has been overlooked. With Qt 6.10, the compiler is able to raise awareness of this issue, hence this PR.

This solves #1346.


kaamui:
Hi @Vekhir,

Thank you for this. I'd like to merge your PR, can you take the asked modifications into account (or debate them if you don't agree) so I can merge it ? Thanks !

Vekhir:
@letsfindaway @kaamui Thanks for the feedback, will push an update shortly!

Vekhir:
@kaamui Pushed the update, compiles and is ready to merge.

PR #1370, opened 2025-11-02
fix: shortcuts while editing text item

### LINKED ISSUE #1367: [Bug] Pasting text with shift-insert into an existing textbox clears all previous text
### Describe the bug
When attempting to add additional text to a textbox using shift-insert (ctrl-v seems to work ok) the existing text in the box is completely replaced but worse than that, it breaks the **undo** for the textbox.

### To Reproduce

1. Create a text box with some text.
2. Copy text from any source
3. Paste the new text using shift-insert.

#### Expected behavior
The new pasted text should be inserted into the textbox at the cursor insertion point taking on whatever text attributes are in place at the insertion point, unless some existing text is already selected.

#### Actual behavior
Any existing text is removed and the only the new pasted text remains.
**In addition undo/ctrl-z does not restore the text prior to the paste command.**

### Additional resources
Don't know if it is pertinent, but I noticed the magnet icon is depressed (as expected) during the "shift" part of the hot-key.

It also appears that when shift-insert is used the action/modification to the textbox does not get added to the undo list (it does work with Ctrl-Z).

### Context

  - OpenBoard version :
      - bug happens with : `1.7.3`
      - didn't check : `any other version`
  - OS : 
      - bug happens with : `Linux`
      - didn't check : `Windows` `macOS` 
  - Installation source : `snap`
  
### Additional context

- OS(es) detailed version (for the ones you reproduced the bug on) 
    - Linux : `Ubuntu 22.04.5`
- number and resolution of monitors : `3 `
  - HDMI-0 connected 3840x2160+0+0
  - DP-1 connected 1920x1080+5760+1781
  - DP-2 connected primary 1920x1080+3840+1781
- if used, manufacturer and model of graphic tablet input device : ` N/A`

letsfindaway:
- While editing a text item, secondary shortcuts like `Shift+Ins` have not been processed properly
- Instead they have been routed to the board controller
- Add `UBApplication::enableEditingShortcuts` to enable/disable global action shortcuts for cut/copy/paste
- Disable these shortcuts while a text item is selected
- Reorder and cleanup includes in `UBGraphicsTextItemDelegate`

Fixes #1367

### DIFF
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -2450,13 +2450,14 @@ void UBBoardController::processMimeData(const QMimeData* pMimeData, const QPoint
         if("" != pMimeData->text()){
             // Sometimes, it is possible to have an URL as text. we check here if it is the case
             QString qsTmp = pMimeData->text().remove(QChar('\0'));
-            if(qsTmp.startsWith("http"))
+
+            if (qsTmp.startsWith("http:") || qsTmp.startsWith("https:"))
+            {
                 downloadURL(QUrl(qsTmp), QString(), pPos);
-            else{
-                if(mActiveScene->selectedItems().count() && mActiveScene->selectedItems().at(0)->type() == UBGraphicsItemType::TextItemType)
-                    dynamic_cast<UBGraphicsTextItem*>(mActiveScene->selectedItems().at(0))->setHtml(pMimeData->text());
-                else
-                    mActiveScene->addTextHtml("", pPos)->setHtml(pMimeData->text());
+            }
+            else
+            {
+                mActiveScene->addTextHtml("", pPos)->setHtml(qsTmp);
             }
         }
         else{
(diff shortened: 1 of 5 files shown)


letsfindaway:
I reverted this PR back to "Draft" because I now think that it is better to coordinate the editing shortcuts with the **focus** state of the text item instead of the **selection** state. Only one text item can have the focus, but multiple might be selected. So I will change this.

My intended target behavior is as follows:

- If a text item has focus, then disable global editing shortcuts, so that they go to the item.
- If no text item has focus, then enable global editing shortcuts, so that they go to the `UBBoardController`.
- In the `UBBoardController` remove the check whether a text item is selected.  When pasting text, we just have two cases:
  - The text is a URL starting with `http:` or `https:`, then start creating a web widget.
  - In all other cases create a new text item.

Note that the latter should happen independent of the selection state of a web widget. The selection state shall never influence paste handling, only the focus state will. For the user, the focus state is indicated by the fact, that there is a text cursor within the text.

letsfindaway:
ready for review with the above mentioned behavior implemented.

PR #1386, opened 2025-12-05
Fix compilation with poppler 25.11

jbruechert:
Patch used here: https://github.com/flathub/ch.openboard.OpenBoard/pull/105

### DIFF
--- a/src/pdf/XPDFRenderer.cpp
+++ b/src/pdf/XPDFRenderer.cpp
@@ -62,7 +62,7 @@ XPDFRenderer::XPDFRenderer(const QString &filename, bool importingFile)
         globalParams->setupBaseFonts(QFile::encodeName(UBPlatformUtils::applicationResourcesDirectory() + "/" + "fonts").data());
     }
 #if POPPLER_VERSION_MAJOR > 22 || (POPPLER_VERSION_MAJOR == 22 && POPPLER_VERSION_MINOR >= 3)
-    mDocument = new PDFDoc(std::make_unique<GooString>(filename.toLocal8Bit()));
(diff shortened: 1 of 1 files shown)


letsfindaway:
Did the `GooString` API change? Or is it more an issue with Qt versions which allow or deny the implicit conversion of a `QByteArray` to a `const char *`?

Anyway doing an explicit conversion as you proposed would be cleaner.

jbruechert:
> Did the `GooString` API change? Or is it more an issue with Qt versions which allow or deny the implicit conversion of a `QByteArray` to a `const char *`?
> 
> Anyway doing an explicit conversion as you proposed would be cleaner.

The compilation issue was introduced with the recent poppler update. As far as I understand, they added a new overload for std::string_view, which made the previous implicit conversion ambigous.

Vekhir:
The poppler constructors are all marked explicit.

The deeper issue is that Qt 6.10 also has conversion operators for both `const char*` and `string_view`, which allow conversions even for explicit constructors. The conversion operator for `const char*` shouldn't be used, instead `constData()` as is proposed here.

We could also go the other way and explicitly cast to `string_view`, which I think is a better choice than `const char*`. Of course, this means a new `#ifdef`.

I added a patch using static cast to `string_view` in #1387

jbruechert:
> We could also go the other way and explicitly cast to `string_view`, which I think is a better choice than `const char*`. Of course, this means a new `#ifdef`.
> 
> I added a patch using static cast to `string_view` in #1387

Ah nice, this saves one `strlen`.

kaamui:
https://github.com/OpenBoard-org/OpenBoard/pull/1387 has been merged

PR #1390, opened 2025-12-11
Fix: some bugs in UBDocument

letsfindaway:
This draft PR is used to make further improvements to `UBDocument` and `UBDocumentToc`. It fixes some errors I found when testing and adds success/failure feedback for the page manipulation functions of `UBDocument`.

I start this with a Draft PR, as I might add more commits when I find more problems or improvements.

The first commit adds a `bool` return value for page manipulation functions in `UBDocument`. It also adds a parameter to suppress the confirmation dialog when deleting pages. This might be useful later for the widget API. Additionally it fixes an issue that thumbnail and assets list have not been created when adding a new page.

The second commit fixes a crash which could occur when deleting the current scene. The function `deleteScene()` in the `UBBoardController` did already handle this problem using the variable `mDeletingSceneIndex`. But this function is no longer used. I now added code to the callback `documentSceneDeleted()` to handle this case. `deleteScene()` is removed, as it is no longer used.

### DIFF
--- a/src/board/UBBoardController.cpp
+++ b/src/board/UBBoardController.cpp
@@ -566,7 +566,9 @@ void UBBoardController::documentSceneDeleted(UBDocument* document, int index)
         if (UBApplication::applicationController->displayMode() == UBApplicationController::Board)
         {
             // directly change scene
+            mDeletingSceneIndex = index;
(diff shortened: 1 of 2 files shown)


letsfindaway:
Ready for merging. I'm not adding more changes to this PR.

PR #1408, opened 2026-02-03
Fix widget name in preferences.ui

Vekhir:
#955 introduced a new preference dialog which includes a new scroll area. The chosen name for it isn't unique however, thus this warning appears during building:
```
[  3%] Automatic MOC and UIC for target openboard
AutoUic: /build/openboard-git/src/OpenBoard/resources/forms/preferences.ui: Warning: The name 'scrollAreaWidgetContents' (QWidget) is already in use, defaulting to 'scrollAreaWidgetContents1'.
```

It's better for us to control the name instead of some default, and adapt the name to the used convention with underscore and number.

### DIFF
--- a/resources/forms/preferences.ui
+++ b/resources/forms/preferences.ui
@@ -1175,7 +1175,7 @@
          <property name="styleSheet">
           <string notr="true">QScrollArea:disabled {border: 1px solid red;}</string>
          </property>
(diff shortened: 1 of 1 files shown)


Vekhir:
Thanks, that was quick

PR #1409, opened 2026-02-03
refactor: use different setContent overload

Vekhir:
4586cc3d32ac0ca5c2d455d4aad87ba0ea51776b added a deprecated use of setContent.

Similar to a5141d84ca1b4737b42133395fe6571945a4e7ca, namespace processing can be specified via ParseOptions instead of a boolean value.


Vekhir:
Thanks for the quick merging!

PR #1413, opened 2026-02-06
fix: screen capture for KDE on Wayland

### LINKED ISSUE #236: [Bug] Screenshot with KDE and Wayland
I just started testing the current `dev` regarding the Wayland features. I have seen that you had to add some changes for GNOME, but those now don't work well with KDE.

The points I have are:

- When you set the `interactive=true` option, a dialog is shown when requesting a screenshot in Desktop mode. On KDE, this dialog is quite limited (and look ugly): You can just select three options:
  - Fullscreen (means all screens)
  - Current screen
  - Current window

  and you can select to include or exclude the mouse cursor, but you have no option to select an arbitrary area.
- Additionally the screenshot contains the KDE dialog. This might be an KDE error, but of course it is not helpful.

So I think for KDE I have to use the non-interactive mode (which always captures all screens) and need the `UBCustomCaptureWindow` to select the area.

It is really bad that the application has to handle different behaviors of GNOME, KDE and probably other desktop environments, but this is how Wayland behaves: You have an API (the desktop portal), but you cannot predict what happens when you invoke it. Sigh.

I'm quite sure we will come across similar things in the future. So should we introduce a common piece of code to determine the desktop environment? There is the `XDG_CURRENT_DESKTOP` environment variable which should tell us.

@kaamui: Comments? More ideas?

letsfindaway:
- the KDE implementation of the `ScreenShot` portal does not allow for user clipping in interactive mode
- add a function to determine the desktop environment to `UBPlatformUtils`
- expose this function as `grabCanCrop()`
- use this function to determine whether user clipping has to be done after the screenshot using the `UBCustomCaptureWindow`
- allow interactive and non-interactive screenshots in `UBDesktopPortal`

**Note:** please see my comments in `UBDesktopPortal::handleScreenshotResponse()` about coordinates of a full screenshot in non-interactive mode. This PR is a Draft until this question is resolved.


letsfindaway:
Added two more commits:

- The `keepOnTop()` function did not work with Plasma 6, because the path to a loaded script changed. Now the path is determined correctly for Plasma 5 and 6.
- The position of the glass pane changed when enabling mirroring. I could observe this in my setup, where the control screen is to the right of the display screen. I fixed this by showing the glass pane (`mTransparentDrawingView` in the `UBDesktopAnnotationController`) full screen. Wayland needs this to position the window to the desired screen.

letsfindaway:
Added additional commit to refactor the desktop portal classes according to our discussion.

I still have this question open:

https://github.com/letsfindaway/OpenBoard/blob/11d76c7be0978584e6c85e8972ea9e8216101714/src/frameworks/linux/UBScreenshotDesktopPortalWrapper.cpp#L119-L124

PR #1434, opened 2026-02-25
build: adapt to poppler 26.02

letsfindaway:
- poppler 26.02 removes an obsolete parameter from the constructor of SplashOutputDev
- conditionally remove this parameter depending on poppler version

### DIFF
--- a/src/pdf/XPDFRenderer.cpp
+++ b/src/pdf/XPDFRenderer.cpp
@@ -235,7 +235,11 @@ QImage* XPDFRenderer::createPDFImageUncached(int pageNumber, qreal xscale, qreal
         if(mSplashUncached)
             delete mSplashUncached;
 
+#if POPPLER_VERSION_MAJOR > 26 || (POPPLER_VERSION_MAJOR == 26 && POPPLER_VERSION_MINOR >= 2)
+        mSplashUncached = new SplashOutputDev(splashModeRGB8, 1, constants::paperColor);
+#else
         mSplashUncached = new SplashOutputDev(splashModeRGB8, 1, false, constants::paperColor);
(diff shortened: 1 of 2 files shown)


kaamui:
Hi @letsfindaway 

into `master` ? Needed directly on 1.7.6 you think ? Cannot be added to 1.7.7 ?

letsfindaway:
> Hi @letsfindaway
> 
> into `master` ? Needed directly on 1.7.6 you think ? Cannot be added to 1.7.7 ?

Let me say: it breaks building of 1.7.6 for Arch and openSUSE Tumbleweed and probably Debian Testing. Of course I can add this as patch then. But it would be nice to have this merged.

kaamui:
I ask because I already published a new release and built my 3 packages based on this master. I can rebuild my packages but changing the tag is an issue. 

If it is not too much complications for you I prefer not to rebuild the app and change the tag.

letsfindaway:
> I ask because I already published a new release and built my 3 packages based on this master. I can rebuild my packages but changing the tag is an issue.

Ok, then I will add this as patch and you may change the target for the pull request.

kaamui:
> > I ask because I already published a new release and built my 3 packages based on this master. I can rebuild my packages but changing the tag is an issue.
> 
> Ok, then I will add this as patch and you may change the target for the pull request.

1.7.7-dev ? ^^

letsfindaway:
> 1.7.7-dev ? ^^

Yes, I also see in my experiments that the patch might have to be adapted. So let's wait anyway with merging

letsfindaway:
The patch works and is now part of the build for openSUSE Tumbleweed.

PR #1472, opened 2026-04-14
Podcast Preferences Dialog

Vekhir:
Currently, the podcast preferences are set with a tiny menu on the podcast palette. This PR replaces that menu with a dedicated dialog window which also shows which settings are applied when choosing a specific recording profile and allows to change the bit rate and frame rate from within OpenBoard.
All previous options and settings are retained. The dialog itself is based on the color preferences dialog.
I think the dialog is self-explanatory. The design can probably be improved (though after some time, it doesn't look as horrible anymore), but the key functionalities are all in place and ready-to-go.

I should note that I only tested on Arch Linux with the CMake community build and Qt 6. I don't expect any breakage, but if you do experience issues, please let me know.
Also, I didn't touch the translations. Any help with that is appreciated.

This PR is preparation for #1232.
Partially solves #991 #106

### DIFF
--- a/resources/darkTheme.qss
+++ b/resources/darkTheme.qss
@@ -1020,6 +1020,106 @@ QDialog#colorPreferencesDialog QDialogButtonBox#colorPreferencesButtonBox QPushB
     font-weight: 600;
 }
 
+QDialog#podcastPreferencesDialog {
+    background-color: #353535;
+}
+
+QDialog#podcastPreferencesDialog QWidget#podcastPreferencesHeader {
+    background-color: #2b2b2b;
+    border: 1px solid #555555;
+    border-radius: 8px;
+}
+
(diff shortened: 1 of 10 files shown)


Vekhir:
Hi @kaamui!
The preferences dialog looks like this (all default settings):
openboard-podcast-preferences-dialog
The recording tab sets the audio recording device with all current modes supported; the frame rate and base bit rate can be set globally.
Via a three tab menu, the desired profile can be selected and shows the applicable settings that will take effect (or are in effect).

The upload tab allows for intranet or youtube upload:
openboard-podcast-preferences-dialog-upload

I think this closely matches the design we discussed in #1232. What do you think?

kaamui:
I think it's amazing work ! I can't look at OpenBoard now but when I'm back on it I'll take a look immediately on your PR !

Thank you a lot for the work !

PR #1474, opened 2026-04-15
Fix Poppler 26.04 build failure

Vekhir:
The code breaks with poppler 26.04. There are no known uses of this function in the code base, and limited testing showed no regressions.


letsfindaway:
This patch can also be applied to `master` to build 1.7.7 on systems with poppler 26.04.

PR #1487, opened 2026-05-08
Add preference to disable automatic pen/eraser switching

### LINKED ISSUE #1432: [Feature Request] Option to disable automatic switch between pen and eraser
**Is your feature request related to a problem? Please describe.**
I am using Openboard with a Promethean Activ Panel.
When I hover my finger or pen above writing, Openboard switches from pen to eraser. 
This is extremely annoying, because it happens a couple of times per lesson, especially when Students write at the board.

**Describe the solution you'd like**
I would like an off-switch for this feature.

SofianElmotiem:
Fixes the annoyance where Openboard auto-switches to the eraser tool when a tablet device (like the Promethean Activ Panel) detects the eraser pointer type. There is now a checkbox in the Pen preferences tab to turn this off. Defaults to on so nothing changes for existing users.

Closes #1432

### DIFF
--- a/resources/forms/preferences.ui
+++ b/resources/forms/preferences.ui
@@ -951,7 +951,7 @@
        <item row="1" column="0" colspan="2">
         <widget class="QCheckBox" name="autoSwitchToEraserCheckBox">
          <property name="text">
-          <string>Automatically switch to eraser when tablet eraser is detected</string>
+          <string>Ignore broken eraser detection for tablets</string>
          </property>
         </widget>
        </item>
--- a/resources/forms/preferences.ui
+++ b/resources/forms/preferences.ui
@@ -948,6 +948,13 @@
          </property>
         </widget>
        </item>
+       <item row="1" column="0" colspan="2">
+        <widget class="QCheckBox" name="autoSwitchToEraserCheckBox">
+         <property name="text">
+          <string>Automatically switch to eraser when tablet eraser is detected</string>
+         </property>
+        </widget>
+       </item>
       </layout>
      </widget>
      <widget class="QWidget" name="markerTab">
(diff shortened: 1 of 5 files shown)


Vekhir:
In my opinion, this seems to be a workaround for broken pens. As such, I'd rather the option make that clear with a text like "Ignore broken eraser detection for tablets", defaulting to off. Similar for the variables.

The proper solution would be a driver fix for the pen to disable this behaviour globally, since other software would behave just like OpenBoard. This is far outside the remit of OpenBoard.

SofianElmotiem:
Thanks for the review! Makes sens
 I'll update the label to make it clear this is a workaround, flip the default to off, and rename the variables accordingly.

Vekhir:
Looks fine now. The commit messages are very structured.

SofianElmotiem:
Your suggestion made the intent much clearer that i need to rethink about it , and let me know if there is anything else to adjust before this can be merged.

Vekhir:
Some people like to add a `Signed-off-by: ` to signify that they as a human are the original authors of the code. It's part of the [Developer Certificate of Origin](https://developercertificate.org/).

What do you think?

SofianElmotiem:
Honsetly did not know it before but for sure happy to add it

Vekhir:
I saw that your account is quite new, so I thought to mention it.
From this point, I don't have anything else to add. Let's see what the others might say.

SofianElmotiem:
thats right it took me some time considering contributing to opensource but i did it in the end. so i am happy to share and learn from the community.

Vekhir:
Welcome to opensource!

abdna94:
It was really interesting to read through this discussion. Glad to see the layout and naming are much clearer now after following Vekhir's suggestions. 

Thanks @SofianElmotiem for setting up the fix and adding the sign-off, learned something new from this PR! Everything looks good to me.

Issue #916, opened 2024-03-19
Good day. How to open OpenBoard in windowed mode in version 1.7.0?


EkaSanatani:
According to version 1.7.0 release note, Openboard now supports windowed mode. Anyone know how to activate / use it. 
Thanks
![image](https://github.com/OpenBoard-org/OpenBoard/assets/162059566/ced98c63-c77b-40e8-aa4e-1341c7e5b646)

letsfindaway:
You can do that by editing the user configuration file:

- Locate the `OpenBoardUser.config` file. Instructions are given here, depending on your operating system: [Configuration file location](https://github.com/OpenBoard-org/OpenBoard/wiki/Configuration#configuration-files-location).
- Add or modify the `RunInWindow` flag:

```ini
[App]
RunInWindow=true
```

You can find more parameters and their meaning in the [Parameters](https://github.com/OpenBoard-org/OpenBoard/wiki/Parameters) documentation

EkaSanatani:
Thank you very much Mr.Martin Winter. 
Works like charm.

For Windows 10:
1.Use Windows search button to search for Notepad.
2. Right click on it and choose  'run as administrator'
3. Using file explorer - click C: then 'Users' folder then 'User' folder
4. Click on 'View' at top bar and click on 'hidden items' to switch it on
5. You should be able to AppData folder in the list, double click on it
6. Then click on 'Local' folder
7. Locate the OpenBoard folder in the list and open it
8. You should be able to see 'OpenBoardUser.config' in the list
9. Right click on it and choose open with Notepad
10. In the Notepad, locate [App]
11. Under [App], copy paste this code 'RunInWindow=true'
12. Close the Notepad and choose save changes
13. Reopen OpenBoard, it should be in windows mode.

Being a non computer guy, I needed to work this steps out through internet search. Listing here for other user to save your time.

Thanks to Mr. Martin Winter again.

VasilisNikolo:
I can  not  save  it.Apperas  the message that  i hav enot the permission

letsfindaway:
> I can not save it.Apperas the message that i hav enot the permission

You should have read and write permissions to the OpenBoard user configuration file. If you experience such problems and want somebody to be able to help you, please provide very detailed information:

- Double-check that you did the steps exactly as listed above.
- What is the exact path and file name of the file you have opened?
- Which program did you use to open the file?
- What is your user name? Does it match the name in the path?
- What is the content of the [App] section of the file?
[...]

PR #1494, opened 2026-05-22
feat: add minimize button and windowed mode for easier window management

riammadber:
## Problem
Teachers cannot minimize OpenBoard during live classes because 
the app runs in borderless fullscreen mode with no minimize button.

## Solution
Added a minimize button to the toolbar that calls showMinimized().
Also added a windowed/fullscreen toggle in Preferences → Display.

## Testing
Tested on Windows 11, OpenBoard v1.7.7


Vekhir:
> Teachers cannot minimize OpenBoard during live classes because the app runs in borderless fullscreen mode with no minimize button.

Are you aware of the Desktop Mode? With it, you can fully access the underlying system without leaving OpenBoard. Maybe that's a better solution to your problem.

sebojolais:
Yes, this feature is already there. 
Have [a look at this ticket](https://github.com/OpenBoard-org/OpenBoard/issues/916).
It is better to open a ticket to discuss the topic before providing a PR.

PR #1496, opened 2026-05-28
feat: Align left sidebar with vertical stylus palette

### LINKED ISSUE #1483: [Feature Request] Stylus bar should not obscure the sidebars.
**Is your feature request related to a problem? Please describe.**
I have the stylus bar on the left side. When I open a sidebar the stylus bar is above the sidebar which makes interacting with the sidebar difficult.

**Describe the solution you'd like**
The sidebar should open beside the stylus bar.

**Describe alternatives you've considered**
The stylus bar could be autohidden, when it collides with the sidebar. (and restored afterwards.)

Thanks.

baetjer:
- Add an optional left offset for the for the left dock palette
- Position the sidebar tab next to the vertical stylus palette
- Update the offset when the stylus palette changes orientation or visibility
- Keep the existing behavior unchanged for horizontal stylus palette

Closes #1483 
References: #1483 

Signed-off-by: claas998, baetjer, SofianElmotiem

### DIFF
--- a/src/board/UBBoardPaletteManager.cpp
+++ b/src/board/UBBoardPaletteManager.cpp
@@ -1008,9 +1008,7 @@ int UBBoardPaletteManager::leftDockPaletteOffset() const
     if (!UBSettings::settings()->appToolBarOrientationVertical->get().toBool())
         return 0;
 
-    constexpr int gap = 0; // optional z. B. 4, wenn du etwas Abstand willst
-
-    return mStylusPalette->geometry().right() + 1 + gap;
+    return mStylusPalette->geometry().right() + 1;
 }
 
 void UBBoardPaletteManager::updateLeftDockPaletteOffset()
--- a/src/board/UBBoardPaletteManager.cpp
+++ b/src/board/UBBoardPaletteManager.cpp
@@ -120,6 +120,7 @@ UBBoardPaletteManager::~UBBoardPaletteManager()
 void UBBoardPaletteManager::initPalettesPosAtStartup()
 {
     mStylusPalette->initPosition();
+    updateLeftDockPaletteOffset();
 }
 
 void UBBoardPaletteManager::setupLayout()
@@ -501,6 +502,7 @@ void UBBoardPaletteManager::containerResized()
         mStylusPalette->move(userLeft, userTop);
         mStylusPalette->adjustSizeAndPosition();
         mStylusPalette->initPosition();
+        updateLeftDockPaletteOffset();
     }
 
     if(mZoomPalette)
@@ -569,6 +571,7 @@ void UBBoardPaletteManager::backgroundPaletteClosed()
 void UBBoardPaletteManager::toggleStylusPalette(bool checked)
 {
     mStylusPalette->setVisible(checked);
+    updateLeftDockPaletteOffset();
 }
 
 
@@ -963,6 +966,8 @@ void UBBoardPaletteManager::changeStylusPaletteOrientation(QVariant var)
 
     connect(mStylusPalette, SIGNAL(stylusToolDoubleClicked(int)), UBApplication::boardController, SLOT(stylusToolDoubleClicked(int)));
     mStylusPalette->setVisible(bVisible); // always show stylus palette at startup
+    mStylusPalette->initPosition();
+    updateLeftDockPaletteOffset();
 }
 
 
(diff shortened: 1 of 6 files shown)


abdna94 (commented):
Looks like a great feature! Just a tiny typo in the PR description's first bullet point, "the" is repeated twice. Other than that, the commits and description look really clean!?

Vekhir:
I don't feel like this is the right solution. For me, the left dock should close with the board frame, and the stylus palette (that's free to move anyway) should adjust to the dock and be kept "inside" the board.
The presented solution leads to visual noise with empty space above and below the vertical stylus palette, and the closed dock having a floating handle in the middle of nowhere.

Thus, the stylus bar should adjust to the dock, not the other way around.

On the commits, the `Signed-off-by` should be on the individual commits for easy reference later on.
Since the second commit only fixes things introduced by the first one, it makes sense to combine them into a single commit. During rebase, there is the option f (fixup) to do just that.

Last note, though it doesn't matter much since the comment got removed; comments should be in English.

Vekhir:
@abdna94 You can ping your teammates, or ask them yourself. I don't think actively waiting will achieve much.

Nevertheless, I'm interested in the considerations and the reasoning behind the decision. I like to add them into my commit messages, so others can follow my thought process. Oftentimes, this is also useful for later, when issues occur or better solutions might arrive.
It's easy to see _what_ the code does, not so much _why_ it's done this way. Hence the commit messages.

Even if sometimes, especially for graphical changes, it requires running the code to get a good understanding of its behaviour.

claas998:
@Vekhir  Thanks for the detailed feedback!

My initial reasoning for this approach was based on the wording of #1483. I understood the requested behavior as keeping the stylus palette in its current left-side position and opening the sidebar next to it, so that the stylus palette no longer overlaps the sidebar.

I agree with you, that keeping the dock/sidebar layout as the stable element and moving or offsetting the vertical stylus palette when the left sidebar is opened seems like the better solution.

Before reworking the implementation, I just want to clarify which behavior would be preferred: should the vertical stylus palette be moved/offset when the left sidebar is opened, or would the alternative mentioned in the issue (automatically hiding the stylus palette while the sidebar is open) be the better approach?

Vekhir:
> should the vertical stylus palette be moved/offset when the left sidebar is opened

This is the preferred behaviour for me. It keeps the full functionality available. It's also consistent with how the stylus palette behaves with respect to the toolbar when the latter is moved to the bottom.

If desired, the stylus palette can always be manually hidden.

src/board/UBBoardPaletteManager.cpp
```diff
@@ -963,8 +970,41 @@ void UBBoardPaletteManager::changeStylusPaletteOrientation(QVariant var)
...
+
+    if(!UBSettings::settings()->appToolBarOrientationVertical->get().toBool())
+        return 0;
+
+    int offset = mLeftPalette->x() + mLeftPalette->width();
+
+    const QRect tabRect = mLeftPalette->getTabPaletteRect();
+
+    if(!tabRect.isNull())
+        offset = qMax(offset, tabRect.x() + tabRect.width());
```
Vekhir: Doesn't `mLeftPalette->x()` and `mLeftPalette->width()` include `tabRect`?\
I'd have thought `qMax(offset, tabRect.x() + tabRect.width()) == offset`

src/board/UBBoardPaletteManager.cpp
```diff
@@ -483,6 +483,12 @@ void UBBoardPaletteManager::connectPalettes()
         }
     }
 
+    if(mLeftPalette)
+    {
+        connect(mLeftPalette, SIGNAL(pageSelectionChangedRequired()),
+                this, SLOT(updateStylusPalettePosition()));
```
Vekhir: This is a nice idea.\
To expand on that, maybe a separate signal could be used instead of `pageSelectionChangedRequired` that could be fired during `UBTabDockPalette::mouseMoveEvent` to provide a smoother change and adapt the stylus palette immediately. At the moment, it only updates at the end, when the mouse is released.

src/gui/UBStylusPalette.cpp
```diff
@@ -104,7 +104,7 @@ void UBStylusPalette::initPosition()
     int parentHeight = pParentW->height();
 
     if(UBSettings::settings()->appToolBarOrientationVertical->get().toBool()){
-        int posX = border();
+        int posX = qMax(border(), verticalLeftOffset);
```
Vekhir: I think adding them instead of using `qMax` is better, as that would create a slight separation to the left dock palette. When the palette is at the bottom, it also has that separation for easier distinction.

Vekhir (changes requested):
Looks good! A few minor questions arose, the general behaviour is what I expected.

qqqqqqqqq9:
Hi,

this is mostly what I had in mind, thanks! 
There just one small glitch. The stylus leftaligns to the right of the handle instead of the left border of the window. Or is this intended?
Thanks. 
P.S. This should be adopted for the right pane, too?

claas998:
Thanks for the review comments.

I adjusted the offset calculation so the stylus palette is now only shifted based on the actual expanded left sidebar width, not the sidebar tab/handle. This removes the extra gap when the sidebar is collapsed and also avoids aligning the stylus palette to the handle instead of the sidebar itself.

I also changed the positioning to keep a small border between the expanded sidebar and the stylus palette, and added a dedicated dock geometry signal so the stylus palette updates continuously while the sidebar is resized, instead of relying on the page selection signal.

I did not extend this to the right pane for now, since this PR is focused on the left sidebar / vertical stylus palette behavior. I think that would be better handled separately if needed.

Vekhir (approved):
Looks good now, just what I had in mind.

Issue #1429, opened 2026-02-24
[Bug] Document filenames and clock widget use UTC instead of local time on Windows


czyrichard:
### Describe the bug

OpenBoard uses UTC time instead of the system's local time on Windows. This causes document filenames (which include the creation date and time) to be incorrect. For users in UTC+8 timezone (e.g., Asia/Taipei), the timestamps in document filenames are 8 hours behind the actual local time, making it difficult to locate documents by their creation date later.

The built-in clock widget ("Horloge") also displays UTC time instead of local time, confirming this is a system-wide time handling issue within OpenBoard.

### To Reproduce

1. Set up a Windows system with a non-UTC timezone (e.g., `Taipei Standard Time`, UTC+08:00). Ensure no `TZ` environment variable is set.
2. Launch OpenBoard.
3. Create a new document.
4. Observe the document filename — the date and time shown are in UTC, not local time.
5. Optionally, add the clock widget ("Horloge") to the board to further confirm: it also displays UTC time instead of local time.

#### Expected behavior

Document filenames should use the system's local time (e.g., UTC+08:00 for Taipei Standard Time). The clock widget should also display local time.

#### Actual behavior

Document filenames use UTC time. For example, a document created at 2026-02-24 10:00 local time (UTC+08:00) gets a filename showing 2026-02-24 02:00 (UTC). The clock widget also shows UTC time, 8 hours behind the actual local time.

### Screenshots

N/A

### Additional resources

**Root cause analysis:**

This appears to be related to Qt WebKit's ICU timezone handling. OpenBoard bundles ICU libraries (`icudt74.dll`, `icuin74.dll`, `icuuc74.dll`), and the ICU layer does not correctly map the Windows timezone name (`Taipei Standard Time`) to the IANA timezone (`Asia/Taipei`). As a result, JavaScript's `new Date()` and likely other internal time functions fall back to UTC.

**Workaround:** Setting the `TZ=Asia/Taipei` environment variable before launching OpenBoard forces the correct timezone and resolves the issue.

### Context

  - reproducibility : `systematic`
  - OpenBoard version :
      - bug happens with : `1.7.5`
  - OS :
      - bug happens with : `Windows 10`
  - Installation source : `OpenBoard github download page`

### Additional context

- OS(es) detailed version (for the ones you reproduced the bug on)
    - Windows : `Windows 10 Pro 10.0.19045`
[...]

PR #1498, opened 2026-05-31
Fix windows utc time

abdna94:
This PR fixes the Windows filename generation issue (#1429). 

The code was previously using UTC time, which caused an issue with filenames on Windows. I've updated it to enforce the system's local time using ".toLocalTime()" and format it correctly.


abdna94:
@Ole Please take a look at this PR and set yourself as the assignee as requested. 

For the rest of the team, please review the changes when you have time!

PR #1508, opened 2026-07-13
Add a toolbar button: print the current board view

mlyle:
Add a board-toolbar print action that fits the visible scene to the selected page and preserves orientation.

Lets users independently choose whether to include the board background color and grid in settings.

Tested on Windows 10 on an old DTEN touch board.  Pretty cool to click twice and get hardcopy.

### DIFF
--- a/resources/etc/OpenBoard.config
+++ b/resources/etc/OpenBoard.config
@@ -128,6 +128,10 @@ PageFormat=A4
 Resolution=300
 UsePDFMerger=true
 
+[Print]
+BackgroundGrid=true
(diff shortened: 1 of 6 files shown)


letsfindaway:
The development of OpenBoard takes place at the `dev` branch. So it would be a better if you base your branches on `dev` and also use `dev` as target of your PRs.

For this PR, the most relevant change is the flexible definition of background patterns. So when rebasing, the code around saving and restoring the background will have to be adapted:

mlyle:
Rebased and retargeted to dev

PR #1509, opened 2026-07-13
Make drawing palettes configurable

mlyle:
Allow user to scale board and desktop drawing palettes from 100 to 200 percent-- else the buttons on my touch board (especially the right arrows on erase) is too small to hit.

Also let users hide individual pen, marker, eraser, selector, and pointer tools.


letsfindaway:
> Allow user to scale board and desktop drawing palettes from 100 to 200 percent-- else the buttons on my touch board (especially the right arrows on erase) is too small to hit.

To trigger the submenus you can already use a long click or touch instead of using the small arrows. Would this also solve your problem?

mlyle:
> To trigger the submenus you can already use a long click or touch instead of using the small arrows. Would this also solve your problem?

Hey, I didn't realize that-- that's better.  But I think my drawing board still benefits from 150% size buttons.

mlyle:
Rebased and targeted to dev

Issue #250, opened 2019-10-16
Multitouch Gestures


Randomizando:
Hi!

This program is excellent, but no have support for multitouch. You have an estimative of time implementation?

kaamui:
Hello,

Due to our limited resources, we will not be able to prioritize on this. The main focus is actually to port OpenBoard to latest qt version, and to release OpenBoard 1.5.4 before the end of the year (which will contain some fixes, essentially).

capri99:
Hi, we've been searching for this feature for a while and have found this code from a Spanish institution implementing multi touch (canvas) in Openboard some time ago.
http://desarrollo.educarex.es/redmine/projects/openboard2/repository/openboard/revisions/d562dcf0abd29f6cc160e6ba00712e6693f5e91d/diff
We're trying to contact the original developer but meanwhile this code can help in the progress of this feature :)
Thanks for your precious time!

PR #1510, opened 2026-07-13
Add multitouch support

mlyle:
Add support for multitouch

* Track state "per touch" rather than globally
* Add config item to disable/enable multitouch, defaulting off
* Add multitouch modes: gesture only, multidraw only, or intelligent auto-selection
* Add strokes to undo when completed

This fixes item #250 and allows multiple people side by side to draw simultaneously.  Tested on Windows 10 on an old DTEN touchboard.


letsfindaway:
See also #963 for a previous attempt to add this feature. I did not do any comparison. So this is just to make you aware of this earlier PR.

mlyle:
Rebased and targeted to dev branch.  

This differs from the previous pull request:

* Adds a config option to enable
* Adds a toolbar button to set mode
* Allows multitouch to do gestures (zoom/pan), multiple people drawing, **or** either depending upon context.  I worked a fair bit on the heuristic and it mostly does the right thing.

mlyle:
Tested and validated on large videoconferencing touchscreens running: Windows 10 IOT and on Ubuntu 26.04LTS Cinnamon.

Occasionally the heuristic loses a stroke when there's a lot of simultaneous activity-- this may be the touchscreen itself rather than the code -- but it's pretty good.  Very usable/nice.

